PR #1, opened 2020-06-02
Storage

oliverlloyd:
## What does this change?
Adds the storage lib from frontend.

## Why should we share this code?
Many codebases use localStorage to persist data but there are some some potential pitfalls around using it that need to be handled. The existing `storage.js` lib in frontend does an excellent job of wrapping `localStorage` (as well as `sessionStorage`) to guard against these pittfalls and also adds additional features such as expiry.

By moving this code up and into @guardian/libs we offer the same error protections and features to all projects.

## Could this code change or will there be new use cases around using it in the future?
This code has reamained unchanged for several years and the same solutions are being repeated in other projects with very little variation (but perhaps not as well as was done in frontend).

## Usage
```typescript
import { storage } from './storage';

const { local, session } = storage;

if (local.isAvailable()) {
  local.set('my-key', {foo: 'my data'});
  local.getRaw('my-key'); // '{"value":{"foo":"my data"}}'
  local.get('my-key'); // {foo: 'my data'}
  local.remove('my-key');

  local.set('expired-key', {foo: 'my data'}, new Date('2001-01-01'));
  local.get('expired-key'); // null
}
```

Please see tests for all use cases

### DIFF
--- a/src/storage.test.ts
+++ b/src/storage.test.ts
@@ -1,0 +1,156 @@
+import { storage } from './storage';
+
+const testStorage = (
+	storageName: 'localStorage' | 'sessionStorage',
+	// eslint-disable-next-line @typescript-eslint/no-explicit-any
+	fn: any,
+) => {
+	const engine = fn(storageName);
+
+	beforeEach(() => {
+		engine.available = true;
+	});
+
+	afterEach(() => {
+		engine.storage.clear();
+	});
+
+	it(`${storageName} - sets availability if it isn't defined`, () => {
+		engine.available = undefined;
+		expect(engine.isAvailable()).toBe(true);
+		expect(engine.available).toBe(true);
+	});
+
+	it(`${storageName} - returns false for availability when there is an error`, () => {
+		const origStorage = engine.storage;
+
+		// not available, if setItem fails
+		engine.available = undefined;
+		engine.storage = {
+			setItem() {
+				throw new Error('Problem!');
+			},
+		};
+		expect(engine.isAvailable()).toBe(false);
+
+		engine.storage = origStorage;
+	});
+
+	it(`${storageName} - returns cached availability when false`, () => {
+		// Even if setItem works, it still returns false from cache
+		engine.available = false;
+		expect(engine.isAvailable()).toBe(false);
+	});
+
+	it(`${storageName} - returns cached availability, even if broken`, () => {
+		const origStorage = engine.storage;
+
+		// Here setItem is explictly broken but it still uses the cached true value
+		engine.available = true;
+		engine.storage = {
+			setItem() {
+				throw new Error('Problem!');
+			},
+		};
+		expect(engine.isAvailable()).toBe(true);
+
+		engine.storage = origStorage;
+	});
+
+	it(`${storageName} - handles strings`, () => {
+		const myString = 'a dog sat on a mat';
+		engine.set('aString', myString);
+		expect(engine.storage.getItem('aString')).toBe(
+			'{"value":"a dog sat on a mat"}',
+		);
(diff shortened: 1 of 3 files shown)


src/storage.test.ts
```diff
@@ -0,0 +1,156 @@
+import { storage } from './storage';
+
+const testStorage = (
+	storageName: 'localStorage' | 'sessionStorage',
+	// eslint-disable-next-line @typescript-eslint/no-explicit-any
+	fn: any,
```
SiAdcock, line 6: Can this be `fn: () => unknown`?

oliverlloyd: I tried unknown and a few more besides but I ended up starting to build a type for the entire storage object, which felt excessive for a test, so I stopped, and did the dirty

src/storage.ts
```diff
@@ -0,0 +1,94 @@
+const store = (type: 'sessionStorage' | 'localStorage') => ({
+	available: undefined as boolean | undefined,
+	storage: window[type],
+	isAvailable(): boolean | undefined {
```
JamieB-gu, line 4: Under what circumstances does this return `undefined`? Isn't it always `boolean`?

src/storage.ts
```diff
@@ -0,0 +1,94 @@
...
+				return null;
+			}
+
+			data = JSON.parse(value);
+
+			if (data === null) {
+				return null;
+			}
+		} catch (e) {
+			this.remove(key);
```
JamieB-gu, line 45: Is it slightly surprising behaviour that an operation named `get` could *delete* data?

src/storage.ts
```diff
@@ -0,0 +1,94 @@
...
+		} catch (err) {
+			this.available = false;
+		}
+
+		return this.available;
+	},
+
+	get(key: string): unknown | null {
+		if (!this.available) {
+			return;
```
JamieB-gu, line 26: ?

```suggestion
			return null;
```

src/storage.ts
oliverlloyd: I agree. To change this though would change the contract we're exposing which could pose problems for frontend. My goal was to try to keep the same api. But yeah, `isAvailable(): boolean {` is a lot nicer

src/storage.ts
oliverlloyd: That. is. weird. Totally agree. I think I see the logic of it, it's deciding that the key is invalid and trying to clean up but there's a strong argument to say it's better to keep the invalid data to help with any debugging. I think it's reasonable to remove this step

src/storage.ts
oliverlloyd: The risk here is there's already code expecting `undefined` - arguably checks should be using falsey but once the api is established it's hard to move away. I might review the code in frontend to see if I can verifiy if this happens or not

src/storage.ts
oliverlloyd: Okay, looking at this more I see that `undefined` does make sense. When `isAvailable === undefined` it means the check has not been made yet, it is litterally undefined. So this is correct.

src/storage.ts
oliverlloyd: I agree it's somewhat odd but I think there's a sense to it.

```
		if (!this.available) {
			return;
		}
```
It only returns `undefined` if the storage mechanism is not available, which is different to the key not being found. I think this is a reasonable attempt to define a richer api instead of just returning falsey. `null` means there was no value for the key you requested. `undefined` means I was not able to make the request but it might still exist.

sndrs:
dont know how i closed this!

sndrs:
oh renamed master to main...

PR #3, opened 2020-09-25
Prepare for publishing

sndrs:
## What does this do?

make this a publishable, barebones package we can start to add things to. attempts to be secure but light-touch (hopefully!), and borrow heavily from things i found useful in https://github.com/guardian/commercial-core

- builds for ES2020
- updated eslint, ts, prettier configs
- added editorconfig, dependabot, codeowners, CI, CD to NPM (once token has been added to repo secrets)
- bumped node version
- updated vscode configs
- added [semantic release](https://semantic-release.gitbook.io/semantic-release/) for CD, and [commitzen](https://github.com/commitizen/cz-cli) and [commitlint](https://commitlint.js.org) to ensure commit messages adhere to [conventional commits](https://www.conventionalcommits.org/)

it's a bit hard to compare main, might be easier to checkout the branch and have a look.

## Why?

I keep thinking of things it might be useful to put in here, so I wanted to make it as easier as possible to do that so we an use it

PR #7, opened 2020-09-29
Storage

sndrs:
## What does this change?

reopens #1 after renaming default branch and updating `main` in #3 

## Why?

its useful!


src/storage.ts
```diff
@@ -0,0 +1,89 @@
...
+
+		return this.available;
+	},
+
+	get(key: string): unknown | null {
+		if (!this.available) {
+			return;
+		}
+
+		let data;
```
SiAdcock, line 29: Is it worth trying to assign a type to this

```ts
type Data = {value: unknown; expires: string | number | Date | undefined} | null | undefined

let data: Data
```

A bit hairy?

src/storage.ts
```diff
@@ -0,0 +1,89 @@
+const store = (type: 'sessionStorage' | 'localStorage') => ({
+	available: undefined as boolean | undefined,
+	storage: window[type],
+	isAvailable(): boolean | undefined {
```
SiAdcock, line 4: When does / should this be called?

SiAdcock:
Looks great! Do you think it's worth putting together some quick docs for it?

sndrs:
closing in favour of #31

PR #11, opened 2020-10-02
Add loadScript

sndrs:
## What does this change?

- add `loadScript`, ported/refactored from [frontend](https://github.com/guardian/frontend/blob/main/static/src/javascripts/lib/load-script.js)
- removes stub files

## Why?

- want to use it outside frontend
- will remove in frontend once this is published

### DIFF
--- a/src/loadScript.test.ts
+++ b/src/loadScript.test.ts
@@ -61,9 +61,11 @@ describe('loadScript', () => {
 		await loadScript(goodURL, {
 			async: true,
 			referrerPolicy: 'no-referrer',
+			className: 'u6ytfiuyoibnpoim',
 		});
-		expect(getTestScript(goodURL)?.async).toBeTruthy();
-		expect(getTestScript(goodURL)?.referrerPolicy).toBe('no-referrer');
+		expect(document.scripts[0]?.async).toBeTruthy();
+		expect(document.scripts[0]?.referrerPolicy).toBe('no-referrer');
+		expect(document.scripts[0]?.className).toBe('u6ytfiuyoibnpoim');
 	});
 
 	it('rejects if the script fails to load', async () => {
--- a/src/loadScript.test.ts
+++ b/src/loadScript.test.ts
@@ -7,31 +7,29 @@ const badURL = 'bad-url';
 // when we detect that a script has been added to the DOM:
 // - if the src is goodURL trigger a load event
 // - if the src is badURL trigger an error event
-const fakeLoader = new MutationObserver((mutations) => {
+new MutationObserver((mutations) => {
 	mutations.forEach((mutation) => {
 		mutation.addedNodes.forEach((addedNode) => {
 			if (addedNode.nodeName === 'SCRIPT') {
-				if ((addedNode as HTMLScriptElement).src.includes(goodURL)) {
+				const addedScript = addedNode as HTMLScriptElement;
+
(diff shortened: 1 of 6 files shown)


src/loadScript.test.ts
```diff
@@ -0,0 +1,66 @@
...
+	});
+
+	it('resolves immediately if a script with matching src is already on page and stops there', async () => {
+		expect(document.scripts).toHaveLength(1);
+		await expect(loadScript(goodURL)).resolves.toMatchObject({
+			type: 'fake-onload-event',
+		});
+		await expect(loadScript(goodURL)).resolves.toBeUndefined();
+		await expect(loadScript(goodURL)).resolves.toBeUndefined();
+		await expect(loadScript(goodURL)).resolves.toBeUndefined();
```
gtrufitt, line 50: Why 3x?

src/loadScript.ts
```diff
@@ -0,0 +1,22 @@
+export const loadScript = (
+	src: string,
+	props?: Omit<Partial<HTMLScriptElement>, 'src' | 'onload' | ' onerror'>,
+): Promise<Event | undefined> =>
+	new Promise((resolve, reject) => {
+		if (document.querySelector(`script[src="${src}"]`)) {
```
gtrufitt, line 6: Do you need to check for `src` being protocol-less or are we happy to just reload if `//` vs `https://`?

gtrufitt: Also, could we see bugs from trying to load a script with attributes but already loaded without? I don't know the use cases of this, but we might want to log something?

src/loadScript.test.ts
```diff
@@ -0,0 +1,66 @@
+import { loadScript } from './loadScript';
+
+const goodURL = 'good-url';
+const badURL = 'bad-url';
+
+const getTestScript = (url: string) =>
+	Array.from(document.scripts).find(({ src }) => src.includes(url));
+
+const fakeLoadEvent = new MutationObserver((mutations) => {
```
gtrufitt, line 9: Could you add a comment to this, as it took me a while to reason with what's happening with MutationObserver

gtrufitt (approved):

src/loadScript.ts
sndrs: hmm, that's a good question. i wonder if we should check for more than just that attribute? e.g. i wonder if you loop through `document.scripts` you get a more fully fledged object?

if we created the script element and only inserted it if no others matched it, it might be more accurate?

src/loadScript.test.ts
sndrs: just a random amount of repetition to make sure it doesn't get added. i'll add a comment

src/loadScript.ts
sndrs: interesting... 
image

sndrs: > Also, could we see bugs from trying to load a script with attributes but already loaded without? I don't know the use cases of this, but we might want to log something?

i think if its already in the page, we should be good to return early?

gtrufitt: > i think if its already in the page, we should be good to return early?

Yeah, i can't think of anything that would modify a script's behaviour based on attributes of the scripts tag, that would break down the line

gtrufitt (approved):
LGTM

PR #15, opened 2020-10-05
Enable semantic release on main

sndrs:
## What does this change?

- enables semantic release on pushes to `main`

## Why?

- so we can use it

### DIFF
--- a/.github/workflows/CI.yml
+++ b/.github/workflows/CI.yml
@@ -120,4 +120,4 @@ jobs:
         env:
           GITHUB_TOKEN: ${{ secrets.RELEASE_TOKEN }}
           NPM_TOKEN: ${{ secrets.NPM_TOKEN }}
-        run: npx semantic-release --dry-run
+        run: npx semantic-release


.github/workflows/CI.yml
```diff
@@ -120,4 +120,4 @@ jobs:
         env:
           GITHUB_TOKEN: ${{ secrets.RELEASE_TOKEN }}
           NPM_TOKEN: ${{ secrets.NPM_TOKEN }}
-        run: npx semantic-release --dry-run
+        run: npx semantic-release
```
gtrufitt, line 123: This is probably all good, but might be sped up by doing a --dev install and using https://github.com/actions/cache to cache node_modules? Not a blocker!

gtrufitt (approved):

.github/workflows/CI.yml
sndrs: yeah that's a good point. i need to update the CI a bit, i'll include in there

PR #16, opened 2020-10-05
Ci improvements

sndrs:
## What does this change?

- pulls the node version from the nvmrc file, rather than using 
- explicitly requires node 14
- uses installed copy of `semantic-release` rather than npx

## Why?

- `nvm install` doesn't actually use the installed version in subsequent steps


.github/workflows/CI.yml
```diff
@@ -22,9 +22,15 @@ jobs:
       - name: Checkout repo
         uses: actions/checkout@v2
 
-      - name: Use Node.js
-        shell: bash -l {0}
-        run: nvm install
+      # based on https://github.com/actions/setup-node/issues/32#issuecomment-539794249
```
gtrufitt, line 25: Interesting!

gtrufitt (approved):

PR #25, opened 2020-10-14
fix: do not commit bumps to package.json in CD

sndrs:
## What does this change?

- stops committing bumps to `package.json` and `CHANGELOG.md` when releasing

## Why?

- we need an access token with admin rights to do it, which the default GitHub user in actions does not have
- they aren't needed – releases will still be tagged in git, and available from NPM and GitHub with the correct version in `package.json`, it just won't be committed to `main`
- messages like https://github.com/guardian/commercial-core/pull/58#issuecomment-708456726 won't come from me!

_nb fix: prefix is just to force a patch release on merge to test it_

### DIFF
--- a/.github/workflows/CI.yml
+++ b/.github/workflows/CI.yml
@@ -151,6 +151,6 @@ jobs:
 
       - name: Release
         env:
-          GITHUB_TOKEN: ${{ secrets.RELEASE_TOKEN }}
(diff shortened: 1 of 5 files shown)


jonathonherbert:
Great to see an approach to this 👌

sndrs:
for context, we've found it worked well in https://github.com/guardian/commercial-core/pull/53 and we're adding it in https://github.com/guardian/consent-management-platform/pull/307

PR #31, opened 2020-10-22
feat: storage

sndrs:
## What does this change?

adds a reworked #1 / #7, with a few differences:

- doesn't include `getRaw` (we should really use the native methods for that)
- uses a class internally
- revamped tests to be more jestish

## Why?

started as a review, then didn't want to update the branch and lose the history


src/storage.ts
```diff
@@ -0,0 +1,119 @@
...
+
+		return this.available;
+	}
+
+	/**
+	 * Retrieve a value from the storage type.
+	 *
+	 * @param key - the name of the value
+	 */
+	get(key: string): unknown {
```
gtrufitt, line 50: I wondered here if there was a json value type to *avoid the `unknown`* but the closest I could find was https://github.com/microsoft/TypeScript/issues/1897#issuecomment-338650717 which tries to define it. I'm not sure if that's better than `unknown` in this case?

gtrufitt (approved):
Beautiful tests 👍

src/storage.test.ts
```diff
@@ -0,0 +1,121 @@
...
+	it(`returns cached availability when false`, () => {
+		// Even if setItem works, it still returns false from cache
+		implementation.__setAvailable(false);
+		expect(implementation.isAvailable()).toBe(false);
+	});
+
+	it(`returns cached availability, even if broken`, () => {
+		// Here setItem is explicitly broken but it still uses the cached true value
+		implementation.__setAvailable(true);
+		setSpy.mockImplementation(undefined);
```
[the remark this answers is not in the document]
sndrs: tbh i'm not sure - this came from oliver's one. i guess it's testing that the cached value is used regardless of what happen – is that useful do you think?

i used `setSpy.mockImplementation(undefined);` to try and mimic a situation where it didn't exist for some reason. the original replaced it with a function that threw an error. maybe that is more realistic?

src/storage.ts
sndrs: any of the suggestions here give me `circularly references itself` errors :(

src/storage.test.ts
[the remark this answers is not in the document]
sndrs: updated to throw instead of assign `undefined`

src/storage.ts
mxdvl: https://github.com/DefinitelyTyped/DefinitelyTyped/blob/master/types/json-schema/index.d.ts

tjmw (approved):
This looks great :+1:

One thought I had was about being able to get rid of having to check `isAvailable()` at the top of each of the public methods. If instead of instantiating directly with `new Storage()` you could use a factory method like `Storage.build()` which does the `isAvailable` check once and either returns a "proper" `Storage` instance (which now doesn't need the checks) or a `NullStorage` instance which has the same interface as `Storage`, but doesn't do anything (`set`, `get` and `remove` are all no-ops). It might make testing easier too since you maybe wouldn't need to peep inside at private attributes. Anyway, just food for thought!

sndrs:
following @tjmw's suggestion i've refactored to only create a `this.storage` ref if we can.

to minimise the effect of this, i've also added a handy technique from @darkwing to only instantiate each storage instance when either type is first accessed. this means if no one ever uses `storage.session`, for example, it never performs the checks.

src/storage.test.ts
```diff
@@ -4,10 +4,14 @@ function functionThatThrowsAnError() {
...
+const LOCAL: StorageName = 'local';
+const SESSION: StorageName = 'session';
+
 describe.each([
-	['local', storage.local, global.localStorage],
-	['session', storage.session, global.sessionStorage],
-])('storage.%s', (title, implementation, native) => {
+	[LOCAL, storage[LOCAL], global.localStorage],
+	[SESSION, storage[SESSION], global.sessionStorage],
+])('storage.%s', (name, implementation, native) => {
```
mxdvl, line 14:
```suggestion
	[storage[LOCAL], global.localStorage],
	[storage[SESSION], global.sessionStorage],
])('%s', (name, implementation, native) => {
```

sndrs: `name` is needed further down for some dynamic imports too

mxdvl: Couldn’t spot it. Ignore me!

sndrs:
after a small journey and discussion with a few other engineers, have refactored this back to check for availablity on demand and not tinker with the `get`ter of the export. basically what @gtrufitt reviewed so will merge if green. thanks everyone!

PR #37, opened 2020-10-26
feat: add getRaw and setRaw to storage

sndrs:
## What does this change?

- adds `getRaw` and `setRaw` methods

## Why?

- they're used a lot, and wrapping them in the checks is useful


SiAdcock (approved):
🦁

PR #39, opened 2020-10-27
perf: lazily create storage types on access

sndrs:
## What does this change?

- refactors storage module to lazily create the storage instance for `local` or `session` on first access

## Why?

the safety check is blocking, so we only want to do it when we know we need it


src/storage.ts
```diff
@@ -1,41 +1,25 @@
...
- * Manages using `localStorage` and `sessionStorage`.
- *
- * Has a few advantages over the native API, including
- * - failing gracefully if storage is not available
- * - you can save and retrieve any JSONable data
- *
- * All methods are available for both `localStorage` and `sessionStorage`.
- */
+class StorageFactory {
+	#storage: Storage | undefined;
```
gtrufitt, line 2:
```suggestion
	// https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Classes/Private_class_fields
	#storage: Storage | undefined;
```

gtrufitt: The reason I think is is useful is because it's newer syntax, so while it may not be relevant forever, I personally hadn't come across it and would find it useful to know what it was called to google it (hash variable javascript is difficult to google....)

gtrufitt (approved):

PR #40, opened 2020-10-27
feat: Add getLocale

sndrs:
## What does this change?

- adds `getLocale` (and a `isString` helper along the way)
- it either returns a current copy (reusing values already set by our apps) or fetches it from the API endpoint

## Why?

- so we can do this in all sorts of places, reusing the lib


src/getLocale.ts
```diff
@@ -0,0 +1,39 @@
+import { isString } from './isString';
+import { storage } from './storage';
+import type { CountryCode } from './types/countries';
+
+const KEY = 'gu.geolocation';
+const URL = 'https://api.nextgen.guardianapps.co.uk/geolocation';
+
+// best guess that we have a valid code, without actually shipping the entire list
+const isValidCountryCode = (country: unknown) =>
+	isString(country) && /^[A-Z]{2}$/.test(country as string);
```
mxdvl, line 10: Could we use the `@guardian/types`—`Country` for this?

Why do we have the whole list in this repo now?

src/isString.test.ts
```diff
@@ -0,0 +1,28 @@
...
+		expect(isString(true)).toBe(false);
+		expect(isString(123)).toBe(false);
+		expect(isString(Symbol('Sym'))).toBe(false);
+		expect(isString(new Object())).toBe(false);
+		expect(isString([])).toBe(false);
+		expect(isString(new Map())).toBe(false);
+		expect(isString(new Set())).toBe(false);
+		expect(isString(new WeakMap())).toBe(false);
+		expect(isString(new WeakSet())).toBe(false);
+		expect(isString(new Date())).toBe(false);
```
mxdvl, line 21: `it.each`?

mxdvl (approved):
I like this, maybe get a few more eyes on it before merging?

src/getLocale.ts
sndrs: ooooh didn't know it was in there – totally

mxdvl: Whoops, I somehow thought we’d put it in there. Ignore this for now. However my comment about having it in this repo still stands. We could test whether it is a `typeof Country`, maybe?

sndrs: this is not checking against the type though, just an approximate test (is a 2 char uppercase string)

mxdvl: Gotcha `isTwoCharacterString` 😉

sndrs: précisément

gtrufitt (approved):

src/isString.ts
```diff
@@ -0,0 +1,3 @@
+export const isString = (_: unknown): boolean => {
+	return Object.prototype.toString.call(_) === '[object String]';
```
SiAdcock, line 2: Why not use `typeof _ === 'string'`. Is it just to catch `typeof new String('abc') === 'object'` ?

src/getLocale.ts
```diff
@@ -0,0 +1,45 @@
...
+
+export const getLocale = async (): Promise<CountryCode | null> => {
+	// eslint-disable-next-line @typescript-eslint/no-unsafe-assignment -- it _is_ any
+	const stored = storage.local.get(KEY);
+
+	// if we've got a locale, return it
+	if (isValidCountryCode(stored)) return stored as CountryCode;
+
+	// use our API to get one
+	try {
```
SiAdcock, line 27: Why is this in a `try`?

src/isString.ts
sndrs: yeah exactly

src/getLocale.ts
sndrs: just in the event the fetch errors in some way. we dont really want to handle it since we can't do anything with it if it does, but this is intended to keep it quiet at least – do you think it's unecessary?

SiAdcock (approved):
My curious questions aside, this looks awesome! 👍

src/getLocale.ts
SiAdcock: I'm just curious really, and to check my understanding. I would expect `fetch` errors would need to be handled in a chained `catch` rather than a `catch` block?

src/isString.ts
sndrs: got it from https://stackoverflow.com/a/17772086 :)

PR #41, opened 2020-10-27
chore: add unpkg export

sndrs:
## What does this change?

- adds an `unpkg` export that links to an ES2015, minified build

## Why?

- enables bundlesize checking via `badgesize.io`
- also adds ability to use `<script src="https://unpkg.com/@guardian/libs/dist/umd/index.min.js"></script>`

### DIFF
--- a/.eslintrc.json
+++ b/.eslintrc.json
@@ -5,6 +5,15 @@
     "browser": true,
     "node": true
   },
-  "extends": ["@guardian/eslint-config-typescript"],
-  "ignorePatterns": ["dist", "coverage"]
+  "parserOptions": {
+    "sourceType": "module"
+  },
+  "extends": ["@guardian/eslint-config"],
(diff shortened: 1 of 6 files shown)


.eslintrc.json
```diff
@@ -5,6 +5,15 @@
     "browser": true,
     "node": true
   },
-  "extends": ["@guardian/eslint-config-typescript"],
-  "ignorePatterns": ["dist", "coverage"]
+  "parserOptions": {
```
gtrufitt, line 8: Not for this PR, but I wonder whether there's benefit of a CLI tool a la create-react-app that hides away the eslint? Not sure how I feel about it

gtrufitt (approved):
Nice

.eslintrc.json
sndrs: yeah i think it def worth looking into. i'm not sure how you get editor integration that way though?

PR #42, opened 2020-10-28
feat: add client-side Logger

mxdvl:
## What does this change?

Adds a team-based client-side `logger`, which enables more logging and prevents over polluting the console.

This has been discussed [here](https://github.com/orgs/guardian/teams/client-side-infra/discussions/13).

Teams who want to use this have to register themselves and add a label colour in `logger.teams.ts`. This label should be AA accessible, having a contrast ratio of 4.5 or more (thanks @buck06191).

## Why?

Many teams want to log a lot of information to the console in production. This allows for logging only when a developer is subscribed to a team, and will help reduce noise in the console.

PR #78, opened 2020-12-07
feat: Add colour contrast checking library

buck06191:
## What does this change?
This adds a library to handle this and flag up if something isn't WebAIM accessible with regards to colour contrast.

## Why?
In order to be as accessible as possible we should make sure that foreground and background colours are chosen such that they meet WebAIM principles and rules. This includes in things like loggers etc., where we might not normally think to include them.

See also: https://github.com/guardian/libs/pull/42#discussion_r537569717


mxdvl:
Would it be worth seeing if we could contribute to one of the existing MDN packages?

mxdvl (commented):
I think this is an excellent addition. There are, however, a few other NPM packages that do the same thing. Should we consider contributing to them?

- https://github.com/misund/get-contrast-ratio
- https://github.com/vinaypillai/ac-colors#colorcontrastratio
- https://github.com/johno/get-contrast

src/types/namedColours.ts
```diff
@@ -0,0 +1,146 @@
+/* eslint-disable @typescript-eslint/naming-convention -- The css names are not camelcase */
+export enum NamedColours {
```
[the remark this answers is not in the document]
buck06191: I've updated this now to have the correct aliases for the different `grey`/`gray`colours.

buck06191:
> I think this is an excellent addition. There are, however, a few other NPM packages that do the same thing. Should we consider contributing to them?
> 
> * https://github.com/misund/get-contrast-ratio
> * https://github.com/vinaypillai/ac-colors#colorcontrastratio
> * https://github.com/johno/get-contrast
> 
> Could you please add a `README` as well?

Added in a README and updated the file name to match the main function.
As for contributing to the existing packages, this sounds like a great idea but I'd also like to stick with using Typescript where possible. Given that none of the linked packages are in Typescript, what would be the best way of going  about contributing to them? There's a reasonable chance that the owners might not want those libraries converting over to Typescript, especially those that are either [significantly bigger than this in scope](https://github.com/vinaypillai/ac-colors#colorcontrastratio) or [those that have split their code across multiple packages](https://github.com/misund/get-contrast-ratio) 

More than happy to move this to one of those packages though if it looks like a better idea

src/isContrastAccessible.ts
```diff
@@ -0,0 +1,47 @@
...
+};
+
+export const isContrastAccessible = (
+	fg: string | ColourName,
+	bg: string | ColourName,
+): { AA: boolean; AAA: boolean } => {
+	const contrast = getContrast(fg, bg);
+	if (contrast) {
+		return {
+			AA: contrast > 3,
```
mxdvl, line 38: Was it not 4.5 for AA?

buck06191:
```suggestion
			AA: contrast > 4.5,
```
Good catch

src/colourConverters.ts
```diff
@@ -0,0 +1,37 @@
...
+				parseInt(result[1], 16),
+				parseInt(result[2], 16),
+				parseInt(result[3], 16),
+		  ]
+		: null;
+};
+
+export const isHexCode = (hex: string): boolean =>
+	/^#?([a-f\d]{2})([a-f\d]{2})([a-f\d]{2})$/i.test(hex) ||
+	/^#?([a-f\d])([a-f\d])([a-f\d])$/i.test(hex);
```
mxdvl, line 30: I think this can be simplified to “either 3 or 6 of [a-60-9]”.

```suggestion
	/^#?[a-f\d]{3,6}$/i.test(hex);
```

Can you tell me why the `#` is optional in this test?

src/isContrastAccessible.README.md
```diff
@@ -0,0 +1,19 @@
...
+```typescript
+import { isContrastFriendly } from '@guardian/libs';
+
+const seagreen = 'seagreen';
+const lightseagreen = 'lightseagreen';
+const blackHex = '#000000';
+
+isContrastFriendly(blackHex, seagreen); // {AA: true, AAA: false}
+isContrastFriendly(blackHex, lightseagreen); // {AA: true, AAA: true}
+isContrastFriendly(blackHex, 'foo bar'); // {AA: false, AAA: false}
```
mxdvl, line 18: Not sure you need variables whose value is identical to their names.

```suggestion
const blackHex = '#000000';

isContrastFriendly(blackHex, 'seagreen'); // {AA: true, AAA: false}
isContrastFriendly(blackHex, 'lightseagreen'); // {AA: true, AAA: true}

// invalid hexadecimal colours or named colours return false
isContrastFriendly(blackHex, 'foo bar'); // {AA: false, AAA: false}
```

You could also use a variable which has more semantic meaning:

```typescript
const mediumConstrat = ["#000", 'seagreen'] // shorthand hex is supported
const highContrast = ["#000000", 'lightseagreen']

isContrastFriendly(...mediumContrast); // {AA: true, AAA: false}
isContrastFriendly(...highContrast); // {AA: true, AAA: true}
```

src/colourConverters.test.ts
```diff
@@ -0,0 +1,50 @@
...
+	hexToRGB,
+	isHexCode,
+	isNamedColour,
+	nameToHex,
+} from './colourConverters';
+
+describe('`isHexCode`', () => {
+	test('should return true given a valid hexcode', () => {
+		expect(isHexCode('#000')).toBe(true);
+		expect(isHexCode('#0101ff')).toBe(true);
```
mxdvl, line 11: If `#` is truly optional, you should test against it!

```suggestion
		expect(isHexCode('#0101ff')).toBe(true);
		expect(isHexCode('123456')).toBe(true);
```

mxdvl (changes requested):
Only step left is to export the relevant functions directly from `index.ts`, and add a small paragraph in the main `README`.

I’ve also added some minor comments on the regular expressions.

src/colourConverters.ts
buck06191: I think this is a typo actually. I adapted it from a different regex so I'll update it. 
For the suggested change: this will allow between 3 and 6 rather than 3 or 6. This would allow something like `#abcd` which I don't think is valid. That's why I split it into two tests originally. 

See below and [this link](https://regex101.com/r/rUypN5/2) for test of this.

image

mxdvl: Ah, right, so this is how you’d do it: `^#([a-f\d]{3}|[a-f\d]{6})$`. I’ve updated [your link][101] with this example. If you prefer the javascript `||`, I still think it would be a little clearer to go with:

```typescript
	/^#[a-f\d]{6}$/i.test(hex) ||
	/^#[a-f\d]{3}$/i.test(hex) ||
```

[101]: https://regex101.com/r/rUypN5/3

src/colourConverters.test.ts
mxdvl: `#` is now enforced, 👍

src/colourConverters.ts
```diff
@@ -0,0 +1,37 @@
+import type { ColourName } from './types/namedColours';
+import { NamedColours } from './types/namedColours';
+
+export const nameToHex = (name: ColourName): string => {
+	const hexcode = NamedColours[name];
+	return hexcode;
+};
+
+export const hexToRGB = (hex: string): number[] | null => {
+	const shorthandRegex = /^#?([a-f\d])([a-f\d])([a-f\d])$/i;
```
mxdvl, line 10: Do you want to enforce the preceding `#` everywhere?
```suggestion
	const shorthandRegex = /^#([a-f\d])([a-f\d])([a-f\d])$/i;
```

src/colourConverters.ts
```diff
@@ -0,0 +1,37 @@
...
+	hex = hex.replace(
+		shorthandRegex,
+		function (m, r: string, g: string, b: string) {
+			return r + r + g + g + b + b;
+		},
+	);
+
+	const result = /^#?([a-f\d]{2})([a-f\d]{2})([a-f\d]{2})$/i.exec(hex);
```
mxdvl, line 18: As above.

```suggestion
	const result = /^#([a-f\d]{2})([a-f\d]{2})([a-f\d]{2})$/i.exec(hex);
```

src/colourConverters.ts
buck06191: @mxdvl It's a good question. I personally find logic in regex very difficult to parse, and so thinking about it a good option might be something like :
```javascript
const longcodeRegex = /^#[a-f\d]{6}$/i.test(hex) 
const shortcodeRegex = /^#[a-f\d]{3}$/i.test(hex)

longcodeRegex.test(hex) || shortcodeRegex.test(hex)
	
```

mxdvl (approved):
Well done, excellent addition.

I’ll raise a PR to use this for the logger labels test, instead of the web API.

sndrs:
thanks for this @buck06191, I can see you've put a lot of time into this.

before going over the code itself, can you provide some insight as to you why you think this ought to be part of `libs` specifically? (as opposed to a good practise we should make sure we use of, which it most definitely is) 

I don't want to overlook the effort that went into this PR, but we need to consider the effort that will also need to go into maintaining it in the future, if we merge it.

do you have a current use case for this? it is something we've reimplemented several times already? i.e is this porting a behaviour we are using in more than one place to a central location for anticipated reuse?

I only ask because while the behaviour is important, adding it here means keeping it up to date, making sure it works with whatever other changes we make to `libs` as a whole etc, which adds maintenance overhead.

i feel like `libs` should be toolkit containing things you are very likely to need at least one of per app. every addition makes it harder to look after, thereby making it harder to be useful. if that's a cost worth paying we should pay it, but we should also be careful about adding to the cost if it won't be outweighed by the benefit.

if this is centralising commonly used colour contrast checks then fair enough. but especially when there are other libraries and services that do the same thing, do we (and probably you @buck06191!) want to accept the maintenance role?

e.g. if we created and published types for https://github.com/misund/passes-wcag to http://definitelytyped.org, or by PR to original package, would we get the same result with less overhead – or even shared overhead – going forward?

gtrufitt:
> I don't know if @gtrufitt @SiAdcock have any thoughts on this?

Yes, I agree - I think this is a great utility (I particularly like the check for luminance!), but I can't see this colour contrast being used inside an app directly within our applications so it doesn't make sense published as a standalone script in the `libs` repo. 

Maybe we need a utilities repo of helpful scripts? Or you could publish this as a separate package / contribute some of your ideas to the wcag one and then use it in the tests? Simon might even have thoughts on how it could be integrated with design system work?

buck06191:
@sndrs @gtrufitt Thanks for the feedback above. I'd not thought too much about whether this belonged in `libs` as a whole before writing it - my main focus was to get a typed library in for the contrast checking. Now that I've had time to think about it, and reading the above, maybe this isn't the right place for it? 

My intention was for this to be a useful script that could be integrated into tests to make sure that when we make design changes or implement something like the `log` library where colour contrast checks like this are necessary. Is there a place for us to put these useful "utility" scripts that might be especially useful in tests? I'm also happy to move to using an existing package but of the ones linked to above they either don't seem to be maintained or are much bigger in scope. Again, happy to look into adding types to one of these instead if that seems a better way of handling this.

It does sound like this addition doesn't belong in this package at least as I'm not sure it's wide enough in scope, but I'll leave this thread open for now to get more feedback

buck06191:
I did another check just now to see if there were any WCAG Typescript options and I found
https://github.com/tmcw/wcag-contrast
https://www.npmjs.com/package/@types/wcag-contrast
which might be a better drop in than creating and publishing this here.

mxdvl:
@buck06191 implemented your suggestion in #167

PR #79, opened 2020-12-08
chore: add team names as types

sndrs:
## What does this change?

adds type for team names, with @JamieB-gu's help

## Why?

image

### DIFF
--- a/src/logger.ts
+++ b/src/logger.ts
@@ -26,8 +26,8 @@ const teams = {
 		font: '#ff7300',
 	},
 	cmp: {
-		background: '#FF1493',
-		font: '#333333',
+		background: '#FF3399',
+		font: '#332020',
 	},
 };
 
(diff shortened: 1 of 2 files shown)


src/logger.ts
```diff
@@ -25,16 +25,21 @@ const teams = {
 		background: '#000000',
 		font: '#ff7300',
 	},
+	cmp: {
+		background: '#FF1493',
+		font: '#333',
```
mxdvl, line 30: Whilst we’re relying on WebAIM’s API, we have to use 6-digit hexes.
```suggestion
		font: '#333333',
```

mxdvl (approved):
Good point. It will force teams to register at the source and release a new version before they can use it, but I think that’s fine.

src/logger.ts
```diff
@@ -65,7 +73,7 @@ export const log: LogCall = (team, ...args) => {
  * Subscribe to a team’s log
  * @param team the team’s unique ID
  */
-const subscribeTo: TeamFunction = (team) => {
+const subscribeTo: TeamFunction = (team: TeamName) => {
```
mxdvl, line 76: If it’s already part of the `TeamFunction` type, it’s a little redundant.

```suggestion
const subscribeTo: TeamFunction = (team) => {
```

PR #94, opened 2021-01-05
fix: Add TX Team

tjmw:
## What does this change?

Adds TX Team.

Screenshot 2021-01-05 at 16 59 58

## Why?

So we can start logging with style 😎 

I found it hard to look at a list of hex codes and pick a colour which wasn't already taken, so I've added brief descriptions as comments next to each entry. Please shout if that's not wanted and I'll remove (I've kept the commit separate).

### DIFF
--- a/src/logger.teams.ts
+++ b/src/logger.teams.ts
@@ -6,23 +6,23 @@
  * */
 export const teams = {
 	common: {
-		background: '#052962',
-		font: '#ffffff',
+		background: '#052962', // dark blue
+		font: '#ffffff', // white
 	},
 	commercial: {
-		background: '#77EEAA',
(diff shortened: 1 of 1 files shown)


sndrs (approved):

mxdvl:
@tjmw: worth noting that in VS code you can get those hex colours to be highlighted!

The only reason I didn’t use named colours is because the contrast ratio tests required them to be in hex form, but that can also be fixed.

image

tjmw:
Thanks for the review and merge @sndrs 👍 

@mxdvl ah good shout about the extension. I wonder if a better approach than adding comments would be adding that extension to `.vscode/extensions.json`.

PR #126, opened 2021-03-05
chore: use actions-setup-node

sndrs:
## What does this change?

uses https://github.com/guardian/actions-setup-node to manage node in github actions

## Why?

simpler config, one place to manage node version


SiAdcock (approved):
🤩

PR #134, opened 2021-03-15
feat: added timeAgo

oliverlloyd:
## What?
Adds the `timeAgo()` function.

Takes an absolute date in epoch format and returns a string representing relative time ago.

### Usage
```typescript
        type Options = {
            verbose?: boolean;
            daysUntilAbsolute?: number;
        };

        timeAgo(twoSecondsAgoAsEpoch) // 'now'
        timeAgo(fiveMinutesAgoAgoAsEpoch) // '5m ago'
        timeAgo(twoDaysAgoAsEpoch) // '2d ago'
        timeAgo(sixDaysAgoAsEpoch, { verbose: true' }) // '6 days ago'
        timeAgo(sixDaysAgoAsEpoch, { daysUntilAbsolute: 4' }) // '12 Mar 2021'
```

## Why?
This function has existed for a long time [in Frontend](https://github.com/guardian/frontend/blob/d7d261ac608a5b1c27a5977a5ef70b978cce91c4/static/src/javascripts/projects/common/modules/ui/relativedates.js#L107) and [also in DCR](https://github.com/guardian/dotcom-rendering/blob/0b3ce92e239b6c2ceab0c04e467e783e04ad4a1d/src/web/lib/dateTime.ts#L111). Hopefully other projects might benefit from using it (@JamieB-gu this kind of function is useful on pages with live content)


oliverlloyd:
[makeRelativeDate in /libs](https://trello.com/c/xgHKJ0SR/2730-makerelativedate-in-libs)

JamieB-gu:
> other projects might benefit from using it

Sounds very sensible. Datetime-type helpers sound like a strong candidate for standardising/sharing. This functionality is currently [duplicated in AR](https://github.com/guardian/apps-rendering/blob/07791376a53aadef5798ecf93a6c47ece4c6c3db/src/date.ts#L53).

gtrufitt:
There's an RFC for date formatting in Source here https://github.com/guardian/source/discussions/503

src/makeRelativeDate.README.md
```diff
@@ -0,0 +1,23 @@
...
+# `makeRelativeDate`
+
+Takes an absolute date in epoch format and returns a string representing relative time ago.
+
+## Props
+```typescript
+type FormatType = 'short' | 'med' | 'long';
+
+type RelativeDateOptions = {
+	notAfter?: number; // Return false if the date ago is past a cutoff
```
mxdvl, line 11: Does the date have to be _before_ this timestamp to return a `string` instead of `false`? Ah, it’s not a timestamp, but rather a number of seconds.

Should this be called `notFurtherThan`?

src/makeRelativeDate.test.ts
```diff
@@ -0,0 +1,105 @@
...
+	it('returns false when past any cutoff value given', () => {
+		const twoYearsAgo = new Date(
+			Date.UTC(2017, 10, 17, 10, 0, 0),
+		).getTime();
+		const oneYearInSeconds = 60 * 60 * 24 * 365;
+		expect(
+			makeRelativeDate(twoYearsAgo, {
+				notAfter: oneYearInSeconds,
+			}),
+		).toBeFalsy();
```
mxdvl, line 93: I think words like `past` makes this a little confusing. One timestamp is absolute and the other relative.

mxdvl (commented):
Standardising date is an excellent idea. I have a few points to raise:

## Language
> There's an RFC for date formatting in Source here guardian/source#503
There’s some good discussion there, and I would reiterate that it’s very puzzling to get a date formatted in another language than English if everything else I’m reading is in English.

## `notAfter`
I find the naming of this prop slightly confusing. Are we expected to pass a timestamp, or a number of seconds ago? I suggest renaming to `notFurtherThan`. I’m also curious to know what the use case for this functionality is.

## Future date
What happens if the date is in the future?

## Long months
Doesn’t `Monday 1 February 2021` feel more appropriate that `Monday 1 Feb 2021` if you have the space? Why truncate the months in long from. Why would the day of the week get special treatment?

src/makeRelativeDate.README.md
```diff
@@ -0,0 +1,23 @@
+
+# `makeRelativeDate`
```
sndrs, line 2: rails has similar helper called [`time_ago_in_words`](https://api.rubyonrails.org/v5.2/classes/ActionView/Helpers/DateHelper.html#method-i-time_ago_in_words), which i think is a great description of what this does. 

what do you think about using that here?

reason i ask is i think the current name is ambiguous:

- what is the relative date it offers me? it could be as simple as `Date.now()` minus the time you pass in
- what does 'make' do here? is it going to return something or produce a side effect?

it's not really 'making' a date, it generates and returns a english sentence describing a duration before now

`timeAgoInWords(time, opts)`, seems to indicate that well, to me?

src/makeRelativeDate.README.md
sndrs: how about `maxAge`?

sndrs: (named after you @mxdvl)

src/makeRelativeDate.ts
```diff
@@ -0,0 +1,164 @@
+type RelativeDateOptions = {
+	notAfter?: number;
+	format?: FormatType;
+	showTime?: boolean;
+};
+
+type FormatType = 'short' | 'med' | 'long';
```
sndrs, line 7:
```suggestion
type Format = 'short' | 'med' | 'long';
```

src/makeRelativeDate.ts
```diff
@@ -0,0 +1,164 @@
...
+type PeriodType = 's' | 'm' | 'h' | 'd';
```
sndrs, line 8:
```suggestion
type Unit = 's' | 'm' | 'h' | 'd';
```

oliverlloyd:
> There's an RFC for date formatting in Source here [guardian/source#503](https://github.com/guardian/source/discussions/503)

Ah, interesting, I wasn't aware of this, thanks.

It is a big area for sure. Maybe we should be looking at an off the shelf solution 🤷‍♂️ maybe what we have in Frontend is already okay as it is. Absolutely worth having that discussion and 👍 to that happening in Source.

I do think we might get benefit from /libs as a stopgap for these types of things ahead of then though? If we want to share code like this we have three options at the moment

1. Put it in Source
2. Put it in a repo like here in /libs
3. Copy paste it and accept the duplication

Option 1. Is ultimately the right place but it's slow, 2 is perhaps not the correct location ultimately but it is a location from where the code can be shared with minimal effort, and 3 is also a valid possibility, it's certainly the easiest and quickest!

src/makeRelativeDate.ts
```diff
@@ -0,0 +1,164 @@
...
+		'Sunday',
+		'Monday',
+		'Tuesday',
+		'Wednesday',
+		'Thursday',
+		'Friday',
+		'Saturday',
+	][day];
+
+const monthAbbr = (month: number): string =>
```
sndrs, line 21: I think month names are more commonly described as 'short' or 'long'

```suggestion
const shortMonth = (month: number): string =>
```

src/makeRelativeDate.ts
```diff
@@ -0,0 +1,164 @@
...
+		'Jun',
+		'Jul',
+		'Aug',
+		'Sep',
+		'Oct',
+		'Nov',
+		'Dec',
+	][month];
+
+const pad = (n: number): number | string => (n < 10 ? `0${n}` : n);
```
sndrs, line 37: can this be replaced with native [`padStart`](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/String/padStart)?

src/makeRelativeDate.ts
```diff
@@ -0,0 +1,164 @@
...
+	yesterday.setDate(today.getDate() - 1);
+	return relative.toDateString() === yesterday.toDateString();
+};
+
+const isWithinPastWeek = (date: Date): boolean => {
+	const weekAgo = new Date().valueOf() - 7 * 24 * 60 * 60 * 1000;
+	return date.valueOf() >= weekAgo;
+};
+
+const isValidDate = (date: Date): boolean => {
```
sndrs, line 61: we have some other `is` methods exported in libs, how do you feel about pulling this out into its own module?

src/makeRelativeDate.ts
```diff
@@ -0,0 +1,164 @@
...
+};
+
+const withTime = (date: Date): string =>
+	` ${date.getHours()}:${pad(date.getMinutes())}`;
+
+export const makeRelativeDate = (
+	epoch: number,
+	opts: RelativeDateOptions = {},
+): false | string => {
+	let delta = 0;
```
sndrs, line 110: `delta` is never reassigned, is there a reason it's not just a `const` on line 121?

src/makeRelativeDate.ts
```diff
@@ -0,0 +1,164 @@
...
+	const then = new Date(Number(epoch));
```
sndrs, line 111: does `epoch` need to be coerced here?

src/makeRelativeDate.ts
```diff
@@ -0,0 +1,164 @@
...
+	const now = new Date();
+	const format = opts.format ?? 'short';
```
sndrs, line 113: not a blocker but it might be easier to pick these up skimming the file if they were defaults set in the function args?

src/makeRelativeDate.ts
```diff
@@ -0,0 +1,164 @@
...
+	const extendedFormatting = opts.format === 'short' || opts.format === 'med';
+
+	if (!isValidDate(then)) {
+		return false;
+	}
+
+	// delta is the number of seconds since the event happened
+	delta = Math.floor((now.getTime() - then.getTime()) / 1000);
```
sndrs, line 121: if you need to describe your variable, i'd say you have the wrong variable name :)

```suggestion
	secondsUntilNow = Math.floor((now.getTime() - then.getTime()) / 1000);
```

src/makeRelativeDate.README.md
oliverlloyd: I actually agree, this name is confusing. However, I'm lifting and shifting this code from DCR which was itself lifted from Frontend and I'm concious that if we ever were to complete the circle and import this into Frontend it would be a lot easier to that work if the naming conventions of such properties had remained the same

src/makeRelativeDate.test.ts
oliverlloyd: Agreed. I'll update this test to be clearer

src/makeRelativeDate.README.md
```diff
@@ -0,0 +1,23 @@
...
+
+Takes an absolute date in epoch format and returns a string representing relative time ago.
+
+## Props
```
sndrs, line 6: we're trying to use a common way of describing methods and props in the readme (largely borrowed from most of Sindre Sorhus's readmes) see https://github.com/guardian/libs/blob/main/src/logger.README.md#methods for a good example

src/makeRelativeDate.README.md
```diff
@@ -0,0 +1,23 @@
...
+
+type RelativeDateOptions = {
+	notAfter?: number; // Return false if the date ago is past a cutoff
+	format?: FormatType; // Change string format (default 'med')
+	showTime?: boolean; // Append time to string
+};
+```
+
+
+## Usage
```
sndrs, line 18: this doesn't quite format correctly, think you're going to need three ticks

oliverlloyd:
> ## Future date
> 
> What happens if the date is in the future?

I think it returns `false` in this situation

```
if (delta < 0) {
```

> ## Long months
> 
> Doesn’t `Monday 1 February 2021` feel more appropriate that `Monday 1 Feb 2021` if you have the space? Why truncate the months in long from. Why would the day of the week get special treatment?

src/makeRelativeDate.ts
```diff
@@ -0,0 +1,164 @@
...
+
+	if (delta < 0) {
+		// Dates in the future are not supported
+		return false;
+	} else if (opts.notAfter && delta > opts.notAfter) {
+		// If the event occured after the cutoff (notAfter) bail out
```
sndrs, line 127:
```suggestion
		// If the event occurred after the cutoff (notAfter) bail out
```

src/makeRelativeDate.README.md
```diff
@@ -0,0 +1,23 @@
+
+# `makeRelativeDate`
+
+Takes an absolute date in epoch format and returns a string representing relative time ago.
```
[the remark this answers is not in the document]
oliverlloyd: Nice, better!

src/makeRelativeDate.README.md
oliverlloyd: I spent some time thinking about this but ending up just going with what it was called before. I didn't think of this name though, that's a really nice api!

src/makeRelativeDate.ts
oliverlloyd: That would be a namespace clash with `Format` which is an existing type used in all the platforms. You could get away with it as it's scoped but it might cause issues with global find and replace actions (which we've been doing a lot of recently!)

src/makeRelativeDate.README.md
oliverlloyd: I went with `maxAge` in honour of @mxdvl

sndrs: in this instance, by paying off a little bit of technical debt we'd get the benefit of a cleaner API going forward (it should be pretty easier to search and replace?). 

the alternative is we invest in future technical debt by carrying over a more confusing API into future projects, which seems like a bad move to me?

(also, in frontend [there are zero instances of `notAfter` actually being used](https://github.com/guardian/frontend/search?q=notAfter)...)

src/makeRelativeDate.ts
sndrs: oooof wow so `Format` is now a reserved word? surely that's a massive footgun awaiting non-experts? (but not one for this PR)

src/makeRelativeDate.ts
oliverlloyd: Fixed in separate commit 👍

src/makeRelativeDate.ts
oliverlloyd: Fixed in a separate commit 👍

src/makeRelativeDate.ts
oliverlloyd: Nice idea. I'll do that in a follow up PR. Carded up here: https://trello.com/c/EvexVKCw/2734-move-isvaliddate-into-its-own-helper-util-in-libs

src/makeRelativeDate.ts
oliverlloyd: 🤷‍♂️ probably not (I'm mostly lifting and shifting code here!) Reading through though I can't see any reason why we can't do this and it's certainly better. I'm surprised I didn't get lint bashed.

Fixed

src/makeRelativeDate.ts
oliverlloyd: I went with `secondsAgo` as that read better further down

src/makeRelativeDate.ts
sndrs: nice one

src/makeRelativeDate.ts
oliverlloyd: Thanks!

src/makeRelativeDate.ts
oliverlloyd: I did breifly check this but it's a little fiddly because of the opts object and this `extendedFormatting` property so I backed off that one

src/makeRelativeDate.ts
oliverlloyd: Nope - good spot 👍

oliverlloyd:
@mxdvl & @sndrs 🙇 amazing reviews - thanks both!

I've gone though each point and hopefully we're in a good place. I was originally trying to maintain the orginal api from Frontend but I think you have both made really good points that justify moving away from that to something better

src/timeAgoInWords.README.md
```diff
@@ -0,0 +1,39 @@
...
+
+Options to control the response
+
+## Examples
+```
+timeAgoInWords(twoDaysAgoAsEpoch) // '2d ago'
+timeAgoInWords(twoDaysAgoAsEpoch, { format: 'long' }) // '2 days ago'
+timeAgoInWords(twoDaysAgoAsEpoch, { format: 'short' }) // '2d'
+timeAgoInWords(threeYearsAgo, { maxAge: 1YearAgo }) // false
+timeAgoInWords(threeDaysAgo, { showTime: true, format: 'long' }) // 'Friday 12 March 2021 10:00'
```
mxdvl, line 38: I’d much prefer `March`, but according to the current code, this is not consistent 😉 

```suggestion
timeAgoInWords(threeDaysAgo, { showTime: true, format: 'long' }) // 'Friday 12 Mar 2021 10:00'
```

gtrufitt: I think this is interesting https://wmich.edu/writing/rules/dates#:~:text=March%2C%20April%2C%20May%2C%20June,.%2C%20Nov.%2C%20Dec. obviously, it would be more ideal to follow our own style guide but there is no reference on where to abbreviate and where not to: 
https://www.theguardian.com/guardian-observer-style-guide-d#sign-in-gate:~:text=dates,sixteen%E2%80%9D%2C%20not%20%E2%80%9Ctwo%20thousand%20and%20sixteen%E2%80%9D

This feels like we should ask Central Prod?

oliverlloyd: I don't see anywhere we're using `long` so it's hard to know the uuse case but I'm inclined to agree that `long` means 'March'

src/makeRelativeDate.README.md
oliverlloyd: If `notAfter` isn't being used in Frontend, and it isn't being used in DCR, and no-one _really_ understands what it's for, we should just get rid of it?

We can revisit if things change but this doesn't seem like an edge case we need to future proof for.

sndrs: Agreed, there's no point shipping code no one uses. Maybe it was once useful ¯\_(ツ)_/¯

src/makeRelativeDate.ts
sndrs: Matches the method name better too

src/makeRelativeDate.README.md
oliverlloyd: Sorry @mxdvl , I've removed your namesake property. Fame is fleeting!

src/timeAgoInWords.ts
```diff
@@ -0,0 +1,154 @@
+type Options = {
+	format?: FormatType;
+	showTime?: boolean;
+};
+
+type FormatType = 'short' | 'med' | 'long';
```
sndrs, line 6: just to raise this again – this shouldn't really come into confict with the content design `Format` here should it? i think it will be scoped to this module?

```suggestion
type Format = 'short' | 'med' | 'long';
```

sndrs (dismissed):
one final comment on `Format` again, and there's some changes i'd like to make to the readme formatting but it's tricky via the github UI and the docs layout isn't documented itself, so i'll do them myself after this merges and cc you (and document the document too)

really good stuff, thanks for sticking with it through the review! 👍

sndrs:
(also, maybe update the PR title the new method name? that's what will be squashed into git history)

src/timeAgoInWords.test.ts
```diff
@@ -70,6 +92,112 @@ describe('timeAgoInWords', () => {
 		).toBe('Friday 15 Nov 2019');
 	});
 
+	it('does not pluralise the unit when the delta is one', () => {
+		const onHourAgo = new Date(Date.UTC(2019, 10, 17, 11, 0, 0)).getTime();
```
mxdvl, line 96: I think there’s a small typo here

```suggestion
const oneHourAgo = new Date(Date.UTC(2019, 10, 17, 11, 0, 0)).getTime();
```

src/timeAgoInWords.test.ts
```diff
@@ -70,6 +92,112 @@ describe('timeAgoInWords', () => {
...
+			timeAgoInWords(threeHoursAgo, {
+				format: 'med',
+			}),
+		).toBe('3h ago');
+		expect(
+			timeAgoInWords(threeHoursAgo, {
+				format: 'long',
+			}),
+		).toBe('3 hours ago');
+	});
```
mxdvl, line 133: the [`it.each` table format][1] could be useful here!

[1]: https://jestjs.io/docs/api#testeachtablename-fn-timeout

```ts
it([
  ['short', '3h'],
  ['med', '3h ago'],
  ['long', '3 hours ago'],
])('only pluralises the unit when format is long', (length, expected) => {
		const threeHoursAgo = new Date(
			Date.UTC(2019, 10, 17, 9, 0, 0),
		).getTime();
		expect(
			timeAgoInWords(threeHoursAgo, {
				format: length,
			}),
		).toBe(expected);
	});
```

mxdvl (approved):
Time flies when you’re having fun!

Two small notes on repeating tests with `test.each` in Jest and a small typo in a variable name.

src/timeAgoInWords.README.md
```diff
@@ -0,0 +1,37 @@
...
+Type: `number`
+
+The date when an event happened in epoch format
+
+`opts`
+
+Type:
+```typescript
+{
+	length?: 'short' | 'med' | 'long'; // Change string length (default 'med')
```
sndrs, line 24: is this because you think `length` is clearer or to remove an instance of `format`?

length to me suggests how many characters i'll get back (`String#length`) rather than the format of the sentence?

oliverlloyd: I changed `format` to `length` because I realised using 'format' here is wrong for the _same_ reasons it is wrong elsewhere and, more importantly, it will be hard to keep Typescript happy if we were to try to import this lib into DCR or AR where `format: Format` means something else.

But yeah, arguably, `length` also suffers from similar problems where it is a common property and appears a lot in code but it's not [a reserved word](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Lexical_grammar#keywords), communicates intent and reads nicely. It is true though that it has a connotation around numbers of charaters. `FormatLength`?

sndrs: ah yeah right of course! 

is it `WordsFormat`?

sndrs: it seems really weird that no two packages can use the same name for a type, surely TS must be able to handle that?

src/timeAgoInWords.test.ts
oliverlloyd: My feeling is with tests, you want them to be as boring and as simple as possible so that if they fail in future the developer who has that error message on his screen can immediately (hopefully) see what the problem is. We want to give our future selfs the easiest experience we can. I worry this pattern could confuse future developers - it certainly confused me! Sure, it's perhaps only 30 seconds effort to google it and work it out, but that's a lot when you're just trying to make the tests pass!

mxdvl: Sure, I just suggest it because it has the added benefit of creating three _distinct_ tests, and you can identify exactly which combination of arguments fails.

Carry on!

sndrs:
are you all good with this now @oliverlloyd?

src/timeAgoInWords.README.md
oliverlloyd: The api is confusing in this area and consequently so is the README. `long` doesn't mean a long absolute date, it means the relative date should be longer

```
short = 1s
med = 1s ago
long = 1 second ago
```

The confusion starts when dates pass 5 and 7 days. We loading more meaning into `long` in this case and returning absolute dates earlier. So with `med` we would return `6d ago` but with `long` this would become absolute.

'Friday 12 Mar 2021' is how all absolute dates look (not just 'long' dates) so any decision should be considered with this in mind

oliverlloyd: Maybe we could simplify this area by removing this special logic? The `long` option is never used, not in Frontend, nor in DCR, so I don't think we need this code:

```
        const extendedFormatting = opts.format === 'short' || opts.format === 'med';

        ... if (isToday(then) || (extendedFormatting && isWithin24Hours(then))) {
		// Hours
		...
	} else if (extendedFormatting && isWithinPastWeek(then)) {
		// Days
		...
	} else  if (isYesterday(then)) {...
```

You could refactor that to:

```
        ... if (isWithin24Hours(then)) {
		// Hours
		...
	} else if (isYesterday(then)) {
		// Yesterday
		...
	} else if (isWithinPastWeek(then)) {
		// Days
		...
	} else ...
```

and get the same result and then `long` just means the relative string is longer, which is still never used but nice to have as an option.

sndrs: > 'Friday 12 Mar 2021' is how all absolute dates look

is this an unexpected result for a method called `timeAgoInWords`? especially is nothing is using it

how about default is medium, with an optional `compact: Boolean` param that returns short, and remove long since it's actually an absolute. 

_or..._

if this is _really_ a guardian date formatter, options something like `compact: Boolean` and `relative: Boolean` so that:

```js
dateTimeInWords(epoch) // 'Friday 12 March 2021 10:00' 
dateTimeInWords(epoch, { compact: true }) // 'Friday 12 Mar 2021 10:00'
dateTimeInWords(epoch, { relative: true }) // '2 days ago'
dateTimeInWords(epoch, { relative: true, compact: true }) // '2d'
```

sndrs: actually, maybe this:
```js
dateTimeInWords(epoch) // 'Friday 12 Mar 2021 10:00' 
```
and 
```js
timeAgoInWords(epoch) // '2 days ago'
timeAgoInWords(epoch, { compact: true }) // '2d'
```

because the point of something like `dateTimeInWords` would be to get a consistent format, and it's already settled on the short version

oliverlloyd: You're right, `timeAgoInWords` isn't being truthful! 🤔 

So the logic, simplified, is:

```
	if (secondsAgo < 0) {
		// Dates in the future are not supported
	} else if (within55Seconds) {
		// Seconds
	} else if (withinTheHour) {
		// Minutes
	} else if (within24hrs) {
		// Hours
	} else if (wasYesterday) {
		// Yesterday
	} else if (within5Days) {
		// Days
	} else if (withinTheWeek) {
		// Include day of week in string - "Friday 15 Nov 2019 13:00"
	}
	return (
		// Over a week, return a simple date - "9 Nov 2019"
	);
```

+1 for using a simpler api (always) but for DCR to keep parity we'd need it to be:

```
dateTimeInWords(epoch) // '2d ago'
dateTimeInWords(epoch, { compact: true }) // '2d'
```

Thus, losing `2 days ago` as an option. Maybe this is fine? We're not using it anywhere so there's no immediate impact, it just means it's not available as a future format (at least not without a bit of work)

oliverlloyd: Arguably, `showTime` could be removed as well? Again, it's not used anywhere outside of the tests.

mxdvl: It seems a little curious to have the day of the week only for day 6 or 7. You wrote earlier that all absolute dates contain the day of the week, but the sample code hints otherwise.

oliverlloyd: I'm working backwards from the orginal code which has some weird edge cases around combinations of length ago + format === long. I'm guess this was battlescar logic left over from living in the wild for a while. I think the intention was to be more descriptive when days of week might be useful and then to fall back to pure dates afterwards?

I'd be in favour of simplifying further as this will keep the code easier to understand and the output shorter?

sndrs: i didn't mean to leave `dateTimeInWords` in, i'm really sorry, that was a hangover from an idea i dismissed – but it does help clear up the absolute thing i think, and @mxdvl's comment speaks to the same issue, which is that after a time it's useful to talk about absolutes and not relative dates. and that time might be different in different places.

how about adding `daysUntilAbsolute: Number = 7`?

also,

>Thus, losing 2 days ago as an option.

begs the question, why return it? if we never returned it then it would be up to the user to add it if useful.

so then:

```js
// epoch is 2 days ago
timeAgo(epoch) // '2 days'
timeAgo(epoch, { compact: true }) // '2d'

// epoch is 10 days ago
timeAgo(epoch) // '9 November 2019'
timeAgo(epoch, { compact: true }) // '9 Nov 2019'
timeAgo(epoch, { daysUntilAbsolute: 30 }) // '10 days'
timeAgo(epoch, { compact: true, daysUntilAbsolute: 30 }) // '10d'
```

oliverlloyd: 😍 `timeAgo` 😍 Brilliant!

Thanks for this suggestion and the other api improvements - really great input! I've implenented the api described above with two exceptions:

1. I kept 'ago'. Both DCR and Frontend expect this by default and I was concerned removing it could lead to '9 Mar 2021 ago` type bugs. I think it's nice to keep the decision about if ago is needed or not in here. 

Sorry, I think my comment above around 'ago'  might have been confusing.

2. I switched `compact` to `extended`. I prefer the _sound_ of `compact`, it reads nice, but the default for Frontend and DCR is to be compact and the longer, extended, format is the optional branch here. So it made more sense to default to compact when no option is given and only use the `extended` format in special situations. Er, if that makes sense!? Basically, all the existing code using this function (or the local version of it) defaults to `2d ago` and so in order to have a drop in replacement, we need `extended` to be the option.

sndrs: > '9 Mar 2021 ago`

oh yeah of course! nice one

sndrs: > it made more sense to default to compact when no option is given and only use the extended format in special situations

👍

sndrs: how do you feel about `verbose` over `extended`? don't want to keep this running forever, so up to you, not a blocker

oliverlloyd: Ha, so I actually searched the internet for 'compact antonym' and got this
Screenshot 2021-03-18 at 17 20 39
So that didn't go well.

I think your idea of 'verbose', which means wordy, is better than 'extended', which implies adding to something. So 👍 

And I'm happy to let this run. Loving the review!

src/timeAgo.test.ts
```diff
@@ -107,13 +106,13 @@ describe('timeAgo', () => {
 
 	it('returns extended format for seconds when this option is given', () => {
 		const tenSecondsAgo = new Date(
```
mxdvl, line 108: Small point:

```suggestion
		const twentySecondsAgo = new Date(
```

src/timeAgo.ts
```diff
@@ -1,16 +1,5 @@
...
-		'Sunday',
-		'Monday',
-		'Tuesday',
-		'Wednesday',
-		'Thursday',
-		'Friday',
-		'Saturday',
-	][day];
-
 const shortMonth = (month: number): string =>
```
mxdvl, line 3: Even though this happens to be exactly the first 3 characters of `longMonth`, I think this makes reading easier than a substring. <3

src/timeAgo.ts
```diff
@@ -86,27 +91,32 @@ export const timeAgo = (
...
 	const now = new Date();
 	const extended = options?.extended;
+	const daysUntilAbsolute = options?.daysUntilAbsolute ?? 7;
 
 	if (!isValidDate(then)) {
 		return false;
 	}
 
 	const secondsAgo = Math.floor((now.getTime() - then.getTime()) / 1000);
+	const veryClose = secondsAgo < 15;
```
mxdvl, line 102: Is it within `5` or `15`? The tests use `5`.

mxdvl (approved):
Great, clean, simple, useful!

Few small comments, which are no dealbreakers.

src/timeAgo.ts
oliverlloyd: Fixed the tests 👍

src/timeAgoInWords.README.md
oliverlloyd: `extended` has been replaced with `verbose` 👍

src/timeAgo.README.md
```diff
@@ -33,6 +33,6 @@ Options to control the response
 timeAgo(twoSecondsAgoAsEpoch) // 'now'
 timeAgo(fiveMinutesAgoAgoAsEpoch) // '5m ago'
 timeAgo(twoDaysAgoAsEpoch) // '2d ago'
-timeAgo(sixDaysAgoAsEpoch, { extended: true' }) // '6 days ago'
+timeAgo(sixDaysAgoAsEpoch, { verbose: true' }) // '6 days ago'
```
mxdvl, line 36:
```suggestion
timeAgo(sixDaysAgoAsEpoch, { verbose: true }) // '6 days ago'
```

src/timeAgo.README.md
```diff
@@ -33,6 +33,6 @@ Options to control the response
...
 timeAgo(sixDaysAgoAsEpoch, { daysUntilAbsolute: 4' }) // '12 Mar 2021'
```
mxdvl:
```suggestion
timeAgo(sixDaysAgoAsEpoch, { verbose: true }) // '6 days ago'
timeAgo(sixDaysAgoAsEpoch, { daysUntilAbsolute: 4 }) // '12 Mar 2021'
```

src/timeAgo.README.md
```diff
@@ -0,0 +1,38 @@
...
+The date when an event happened in epoch format
+
+`opts`
+
+Type:
+```typescript
+{
+	verbose?: `boolean`; // Return a longer, more descriptive string when true
+    daysUntilAbsolute?; `number`  // The cutoff for when dates are returned in absolute format
+}
```
mxdvl, line 26:
```suggestion
```typescript
{
	verbose?: boolean, // Return a longer, more descriptive string when true
    daysUntilAbsolute?: number,  // The cutoff for when dates are returned in absolute format
}
```

src/timeAgo.README.md
```diff
@@ -0,0 +1,38 @@
...
+Options to control the response
+
+## Examples
+```
+timeAgo(twoSecondsAgoAsEpoch) // 'now'
+timeAgo(fiveMinutesAgoAgoAsEpoch) // '5m ago'
+timeAgo(twoDaysAgoAsEpoch) // '2d ago'
+timeAgo(sixDaysAgoAsEpoch, { verbose: true' }) // '6 days ago'
+timeAgo(sixDaysAgoAsEpoch, { daysUntilAbsolute: 4' }) // '12 Mar 2021'
+```
```
mxdvl, line 38:
```suggestion
```ts
timeAgo(twoSecondsAgoAsEpoch) // 'now'
timeAgo(fiveMinutesAgoAgoAsEpoch) // '5m ago'
timeAgo(twoDaysAgoAsEpoch) // '2d ago'
timeAgo(sixDaysAgoAsEpoch, { verbose: true }) // '6 days ago'
timeAgo(sixDaysAgoAsEpoch, { daysUntilAbsolute: 4 }) // '12 Mar 2021'
```
```

**Review by mxdvl — APPROVED** [review]
🕥 📆

File: src/timeAgo.README.md [hunk 29]
```diff
@@ -0,0 +1,37 @@
...
+
+The date when an event happened in epoch format
+
+`opts`
+
+Type:
+```typescript
+type Opts = {
+    verbose?: boolean, // Return a longer, more descriptive string when true
+    daysUntilAbsolute?: number,  // The cutoff for when dates are returned in absolute format
```
sndrs, line 25: should we add something above explaining that this can happen?

sndrs, line 25: is the default 7 still? worth adding here

src/timeAgo.README.md
```diff
@@ -0,0 +1,37 @@
...
+}
+```
+
+Options to control the response
+
+## Examples
+```ts
+timeAgo(twoSecondsAgoAsEpoch) // 'now'
+timeAgo(fiveMinutesAgoAgoAsEpoch) // '5m ago'
```
sndrs, line 34:
```suggestion
timeAgo(fiveMinutesAgoAgoAgoAsEpoch) // '5m ago'
```

src/timeAgo.README.md
oliverlloyd: Explaining what can happen? (Sorry, sometimes the highlight feature in Github, whilst great, can be misleading)

sndrs:
are we there with this now, do you think @oliverlloyd?

oliverlloyd:
> are we there with this now, do you think @oliverlloyd?

I'm happy 👍 

Again, thanks both for giving this one so much time and thought, really amazing input. I'm keen to import this new and improved `dateAgo` function into DCR!

mxdvl (approved):

sndrs (approved):

PR #139, opened 2021-03-22
test: improve tests a little

sndrs:
## What does this change?

- prevents jest looking for tests in `dist`
- stops the `logger` modules tests actually logging

## Why?

- doesn't run pointless tests
- keeps the output cleaner


mxdvl (approved):
Clean!

jamie-lynch (approved):
👍

PR #157, opened 2021-04-16
build: use @guardian/pkgu to build it

sndrs:
## What does this change?

strips out the build step in favour of using [@guardian/pkgu](https://github.com/guardian/pkgu)

## Why?

testing the standardised build process


SiAdcock (dismissed):
Wow, this is so tidy! ♻️

package.json
```diff
@@ -4,38 +4,30 @@
...
+  "types": "dist/types/index.d.ts",
+  "files": ["dist", "src"],
   "scripts": {
-    "build": "npm-run-all clean --parallel compile:*",
+    "prebuild": "yarn clean",
+    "build": "pkgu build",
     "clean": "rm -rf dist",
-    "compile:common": "tsc --project ./tsconfig.build.json --outDir ./dist/cjs --module CommonJS",
-    "compile:esm": "tsc --project ./tsconfig.build.json --outDir ./dist/esm",
-    "compile:umd": "rollup -c",
```
SiAdcock, line 28: Minor: are we dropping UMD support?

sndrs: i think so? it's not in the recs, do we need it? or alternatively, should we discourage it?

PR #161, opened 2021-04-20
chore(deps): Bump deps 2021.04.20

mxdvl:
## What does this change?

Bumps dependencies, and uses `^` for package versions, to allow @dependabot to only fix `yarn.lock` and keep `package.json` as-is.

I also ran `yarn upgrade`.

## Why?

Keep the robots happy.


package.json
```diff
@@ -40,14 +40,14 @@
     "@rollup/plugin-typescript": "8.2.1",
     "@semantic-release/github": "7.2.0",
     "@types/jest": "26.0.22",
-    "@types/node-fetch": "2.5.8",
-    "@typescript-eslint/eslint-plugin": "4.19.0",
-    "@typescript-eslint/parser": "4.19.0",
+    "@types/node-fetch": "^2.5.10",
```
mxdvl, line 43: Using `^` here means that dependabot PRs will not touch package.json at all, which I guess is a nice bonus? It also means less conflicts when cherry-picking commits.

Here’s a useful command to find all open dependabot PRs.

```
git log --all --oneline  --author="dependabot"
```

You may need to prune first:

```
git remote prune origin
```

mxdvl:
Superceded by #173

PR #162, opened 2021-04-20
feat: Get locale from cookie 

ioanna0:
## What does this change?

- Adds cookie module 
- Change locale logic to get the country code from newly [GU_geo_country](https://github.com/guardian/fastly-edge-cache/pull/906) cookie
- Return overridden geolocation if it exist to allow RR team test their banners in different countries
- Call API as fallback when cookie is not present and set its value to cookie
- Frontend PR: https://github.com/guardian/frontend/pull/23739

## Why?
Faster & more accurate geolocation

### DIFF
--- a/src/cookies.README.md
+++ b/src/cookies.README.md
@@ -107,7 +107,7 @@ getCookie({name:'GU_geo_country'}); //GB
 getCookie({name:'GU_geo_country', shouldMemoize: true}); //GB
 ```
 
-## `removeCookie(name)`
+## `removeCookie({name, currentDomainOnly?})`
 
 Returns: `void`
 
@@ -119,8 +119,14 @@ Type: `string`
 
 Name of the stored cookie to remove.
 
+#### `currentDomainOnly`
+
+Type: `boolean`
+
+Set to true if it's a cookie for current domain only, defaults to false
 ### Example
 
 ```js
-removeCookie('GU_geo_country');
+removeCookie({name:'GU_geo_country'});
+removeCookie({name:'GU_geo_country', currentDomainOnly: true});
 ```
--- a/src/cookies.README.md
+++ b/src/cookies.README.md
@@ -15,9 +15,9 @@ import {
 
 ## Methods
 
--   [`setCookie(name, value, daysToLive, isCrossSubdomain, shouldMemoize)`](#setCookie)
+-   [`setCookie(name, value, daysToLive, isCrossSubdomain)`](#setCookie)
 -   [`setSessionCookie(name, value)`](#setSessionCookie)
--   [`getCookie(name)`](#getCookie)
+-   [`getCookie(name, shouldMemoize)`](#getCookie)
 -   [`removeCookie(name)`](#removeCookie)
 
 ## `setCookie`
@@ -50,12 +50,6 @@ Type: `boolean`<br>
 
 Set this true if the cookie is cross subdomain.
 
-#### `shouldMemoize?`
-
-Type: `boolean`<br>
-
-When this is set to true it will keep the cookie in memory to avoid fetching more than once.
-
 ### Example
 
 ```js
@@ -83,12 +77,6 @@ Type: `string`<br>
 
 Value of the cookie.
 
-#### `shouldMemoize?`
-
-Type: `boolean`<br>
-
-When this is set to true it will keep the cookie in memory to avoid fetching more than once.
-
 ### Example
 
 ```js
@@ -105,6 +93,13 @@ Returns: `cookie` value if it exists or `null`
 Type: `string`
 
 Name of the cookie to retrieve.
+
+
+#### `shouldMemoize?`
+
+Type: `boolean`<br>
+
+When this is set to true it will keep the cookie in memory to avoid fetching more than once.
 
 
 ### Example
--- a/src/cookies.README.md
+++ b/src/cookies.README.md
@@ -1,0 +1,13 @@
+# `getCookie(name)`
+
+Returns: `string` or `null`
+
+Returns the content of a cookie or `null` if the cookie does not exist.
+
+## Example
+
+```js
+import { getCookie } from '@guardian/libs';
+
+getCookie('GU_geo_country'); // 'GB'
+```
--- a/src/cookies.README.md
+++ b/src/cookies.README.md
@@ -15,12 +15,12 @@ import {
 
 ## Methods
 
--   [`setCookie(name, value, daysToLive, isCrossSubdomain)`](#setCookie)
+-   [`setCookie({name, value, daysToLive, isCrossSubdomain})`](#setCookie)
 -   [`setSessionCookie(name, value)`](#setSessionCookie)
 -   [`getCookie(name, shouldMemoize)`](#getCookie)
 -   [`removeCookie(name)`](#removeCookie)
 
-## `setCookie`
+## `setCookie({name, value, daysToLive?, isCrossSubdomain?})`
 
 Returns: `void`
 
@@ -53,13 +53,12 @@ Set this true if the cookie is cross subdomain.
 ### Example
 
 ```js
-setCookie('GU_country_code', 'GB')
-setCookie('GU_country_code', 'GB', 7)
-setCookie('GU_country_code', 'GB', 7, true)
-setCookie('GU_country_code', 'GB', 7, false, true)
+setCookie({name:'GU_country_code', value:'GB'})
+setCookie({name:'GU_country_code', value:'GB', daysToLive: 7})
+setCookie({name:'GU_country_code', value:'GB', daysToLive: 7, isCrossSubdomain: true})
 ```
 
-## `setSessionCookie`
+## `setSessionCookie(name, value)`
 
 Returns: `void`
 
@@ -84,7 +83,7 @@ setSessionCookie('GU_country_code', 'GB')
 setSessionCookie('GU_country_code', 'GB', true)
 ```
 
-## `getCookie`
+## `getCookie(name, shouldMemoize?)`
 
 Returns: `cookie` value if it exists or `null`
 
@@ -108,7 +107,7 @@ When this is set to true it will keep the cookie in memory to avoid fetching mor
 getCookie('GU_geo_country'); //GB
 ```
 
-## `removeCookie`
+## `removeCookie(name)`
 
 Returns: `void`
 
--- a/src/cookies.README.md
+++ b/src/cookies.README.md
@@ -2,18 +2,61 @@
 
 Robust API over `document.cookie`.
 
-### Example
+### Usage
 
 ```js
-import { cookies } from '@guardian/libs';
-
+import {
+    getCookie,
+    removeCookie,
+    setCookie,
+    setSessionCookie
+ } from '@guardian/libs';
 ```
 
 ## Methods
 
+-   [`setCookie(name, value, daysToLive, isCrossSubdomain)`](#setCookie)
 -   [`setSessionCookie(name, value)`](#setSessionCookie)
 -   [`getCookie(name)`](#getCookie)
--   [`cleanUp(names)`](#cleanUp)
+-   [`removeCookie(name)`](#removeCookie)
+
+## `setCookie`
+
+Returns: `void`
+
+Sets a cookie taking a name and value, optional daysToLive and optional isCrossSubdomain flag.
+
+#### `name`
+
+Type: `string`
+
+Name of the cookie.
+
+#### `value`
+
+Type: `string`<br>
+
+Value of the cookie.
+
+#### `daysToLive?`
+
+Type: `number`
+
+Days you would like this cookie to live for.
+
+#### `isCrossSubdomain?`
+
+Type: `boolean`<br>
+
+Set this true if the cookie is cross subdomain.
+
+### Example
+
+```js
+setSessionCookie('GU_country_code', 'GB')
+setSessionCookie('GU_country_code', 'GB', 7)
+setSessionCookie('GU_country_code', 'GB', 7, true)
+```
 
 ## `setSessionCookie`
 
@@ -56,20 +99,20 @@ Name of the cookie to retrieve.
 getCookie('GU_geo_country'); //GB
 ```
 
-## `cleanUp`
+## `removeCookie`
 
 Returns: `void`
 
-Removes a list of cookies.
+Removes a cookie.
 
-#### `[names]`
+#### `names`
 
-Type: `string[]`
+Type: `string`
 
-Names of the stored cookies to remove.
+Name of the stored cookie to remove.
 
 ### Example
 
 ```js
-cleanUp('GU_geo_country');
+removeCookie('GU_geo_country');
 ```
--- a/src/cookies.README.md
+++ b/src/cookies.README.md
@@ -16,15 +16,15 @@ import {
 ## Methods
 
 -   [`setCookie({name, value, daysToLive, isCrossSubdomain})`](#setCookie)
--   [`setSessionCookie(name, value)`](#setSessionCookie)
--   [`getCookie(name, shouldMemoize)`](#getCookie)
+-   [`setSessionCookie({name, value})`](#setSessionCookie)
+-   [`getCookie({name, shouldMemoize})`](#getCookie)
 -   [`removeCookie(name)`](#removeCookie)
 
 ## `setCookie({name, value, daysToLive?, isCrossSubdomain?})`
 
 Returns: `void`
 
-Sets a cookie taking a name and value, optional daysToLive and optional isCrossSubdomain flag.
+Sets a cookie taking a config object with name and value, optional daysToLive and optional isCrossSubdomain flag.
 
 #### `name`
 
@@ -58,11 +58,11 @@ setCookie({name:'GU_country_code', value:'GB', daysToLive: 7})
 setCookie({name:'GU_country_code', value:'GB', daysToLive: 7, isCrossSubdomain: true})
 ```
 
-## `setSessionCookie(name, value)`
+## `setSessionCookie({name, value})`
 
 Returns: `void`
 
-Sets a session cookie (no expiry date) taking a name and value.
+Sets a session cookie (no expiry date) taking a config object with name and value.
 
 #### `name`
 
@@ -79,13 +79,12 @@ Value of the cookie.
 ### Example
 
 ```js
-setSessionCookie('GU_country_code', 'GB')
-setSessionCookie('GU_country_code', 'GB', true)
+setSessionCookie({name:'GU_country_code', value: 'GB'})
 ```
 
-## `getCookie(name, shouldMemoize?)`
+## `getCookie({name, shouldMemoize?})`
 
-Returns: `cookie` value if it exists or `null`
+Returns: `cookie` value if it exists or `null`. Takes a config object with name and shouldMemoize params
 
 #### `name`
 
@@ -104,7 +103,8 @@ When this is set to true it will keep the cookie in memory to avoid fetching mor
 ### Example
 
 ```js
-getCookie('GU_geo_country'); //GB
+getCookie({name:'GU_geo_country'}); //GB
+getCookie({name:'GU_geo_country', shouldMemoize: true}); //GB
 ```
 
 ## `removeCookie(name)`
--- a/src/cookies.README.md
+++ b/src/cookies.README.md
@@ -15,7 +15,7 @@ import {
 
 ## Methods
 
--   [`setCookie(name, value, daysToLive, isCrossSubdomain)`](#setCookie)
+-   [`setCookie(name, value, daysToLive, isCrossSubdomain, shouldMemoize)`](#setCookie)
 -   [`setSessionCookie(name, value)`](#setSessionCookie)
 -   [`getCookie(name)`](#getCookie)
 -   [`removeCookie(name)`](#removeCookie)
@@ -50,12 +50,19 @@ Type: `boolean`<br>
 
 Set this true if the cookie is cross subdomain.
 
+#### `shouldMemoize?`
+
+Type: `boolean`<br>
+
+When this is set to true it will keep the cookie in memory to avoid fetching more than once.
+
 ### Example
 
 ```js
-setSessionCookie('GU_country_code', 'GB')
-setSessionCookie('GU_country_code', 'GB', 7)
-setSessionCookie('GU_country_code', 'GB', 7, true)
+setCookie('GU_country_code', 'GB')
+setCookie('GU_country_code', 'GB', 7)
+setCookie('GU_country_code', 'GB', 7, true)
+setCookie('GU_country_code', 'GB', 7, false, true)
 ```
 
 ## `setSessionCookie`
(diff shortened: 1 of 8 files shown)


src/getLocale.ts
```diff
@@ -23,6 +24,12 @@ export const __resetCachedValue = (): void => (locale = void 0);
  */
 
 export const getLocale = async (): Promise<CountryCode | null> => {
+	// return locale from cookie if it exists
+	const fromCookie = getCookie('GU_geo_country');
+
+	if (fromCookie) return fromCookie as CountryCode;
```
mxdvl, line 30: I think we’d want to do a few things here to make this more robust:

1. check whether `isValidCountryCode`
2. move this inside `try / catch`, otherwise we risk having out of sync cookies, memoized variable and local storage.

ioanna0: > 1. check whether isValidCountryCode

I added the check although I think this check is really unnecessary since fastly already returns a valid [Two-letter ISO 3166-1 alpha-2 country code for the country. ](https://developer.fastly.com/reference/vcl/variables/geolocation/client-geo-country-code/)


> 2. move this inside try / catch, otherwise we risk having out of sync cookies, memoized variable and local storage.

Why is that? this new cookie will not be cached at all since it will always be returned from fastly on page load time. if this is not the case (e.g localhost or some domain that doesn't go through fastly) the cookie will `null` and then we will continue with that we were doing before - fetch country code from geolocation endpoint and cache it.

mxdvl: If we’re memoizing the `locale`, so that it can be accessed without needing an extra check (cookie or local storage), it’s worth leveraging that benefit.

I think that the reading from the cookie should come in place of a `fetch` call to the API, but not as a short-circuit upstream. Otherwise, it would be impossible to force a user into a locale on prod, as the cookie override it every time.

Trying to diagram it, I find `A` chain is easier to reason about than `B`.

```
A. let locale? → localStorage? → cookie? → API!

B. cookie? → let locale? → localStorage? → API!
```

ioanna0: Agreed with memoizing it 👍
I've changed it to:
```
C. locale? → cookie? (If yes set to locale)→ localStorage? (If yes set to locale) → API! (Set response to local & localStorage)
```
Does that makes sense?
In prod we shouldn't touch localStorage or API I think.

mxdvl (dismissed):
🗺️

ioanna0:
@guardian/client-side-infra can you please review this? 🙏 I think I need 1 more approval before merging

src/getLocale.ts
ioanna0: I had a re-think of this and I think keeping localStorage in local could lead to confusion. It would be better to keep the same storage as production - i.e cookie, so thinking of changing to that?

`locale? → cookie? (If yes set to locale)→ API! (Set response to cookie)`

src/cookies.README.md
```diff
@@ -1,13 +1,76 @@
...
+
+## `setSessionCookie`
+
+Returns: `void`
+
+Sets a session cookie (no expiry date) taking a name and value.
+
+#### `name`
+
+Type: `string`<br>
```
mxdvl, line 26: I think you can get rid of this 😉 

```suggestion
Type: `string`
```

src/cookies.ts
```diff
@@ -0,0 +1,73 @@
...
+
+	// Remove cookie, implicitly using the document's domain.
+	document.cookie = `${name}=;${path}${expires}`;
+	if (!currentDomainOnly) {
+		// also remove from the short domain
+		document.cookie = `${name}=;${path}${expires} domain=${getShortDomain()};`;
+	}
+};
+
+export const setSessionCookie = (name: string, value: string) => {
```
mxdvl:
```suggestion
export const setSessionCookie = (name: string, value: string): void => {
```

mxdvl (approved):
LGTM, with some small comments.

I’ll let @guardian/client-side-infra give the last +1

ioanna0:
> ioanna0 dismissed mxdvl’s stale review via e3a8e17 27 seconds ago
^ That's automated 🙄

mxdvl (dismissed):

ioanna0:
@gtrufitt and/or @guardian/client-side-infra can you review please?

arelra (dismissed):

src/cookies.README.md
```diff
@@ -0,0 +1,75 @@
+# `cookies`
+
+Robust API over `document.cookie`.
+
+### Example
+
+```js
+import { cookies } from '@guardian/libs';
```
sndrs, line 8: i think this example import is wrong?

src/cookies.test.ts
```diff
@@ -0,0 +1,67 @@
...
...false%22%2C%222001830212%22%3A%22none%22%2C%222004900216%22%3A%22referral%22%2C%222005870214%22%3A%22gc%22%7D; optimizelyBuckets=%7B%7D; s_pers=%20s_fid%3D7BD36E2BEA3FA475-209CB26F08488520%7C1473689665518%3B%20s_daily_habit%3D16315%252C16316%252C16317%252C16318%252C16321%252C16323%252C16325%7C1568297665542%3B%20s_lv%3D1413290266527%7C1507898266527%3B%20s_prev_prop9%3Dno%2520value%7C1417605835703%3B%20s_nr%3D1417605906076-Repeat%7C1449141906076%3B; s_sess=%20s_sv_sid%3D1337774611744%3B%20s_campaign%3DEMCNEWEML6619I2%3B%20s_ppv%3D-%252C17%252C17%252C695%3B%20s_cc%3Dtrue%3B%20s_visit%3D1%3B%20s_sq%3Dguardiangu-network%253D%252526pid%25253DGFE%2525253Anews%2525253AArticle%2525253A10-years-bullying-data%252526pidt%25253D1%252526oid%25253Dhttp%2525253A%2525252F%2525252Fwww.theguardian.com%2525252Fpreference%2525252Fplatform%2525252Fclassic%252…%252525252F2013%252525252Fmay%252525252F23%252525252F1%252526ot%25253DA%3B; s_sq=%5B%5BB%5D%5D; GU_VIEW=responsive; s_cc=true; bwid=kvdn_943h8SSC6h3R5hyJ31g; s_fid=470C3252CA9A4650-35C6B54A21985FBA; s_daily_habit=16406%2C16407%2C16408; s_lv=1417705099834; s_nr=1417705099835-Repeat; s_vi=[CS]v1|29AB343C05013259-4000014980021C3D[CE]; cto2_guardian=';
```
sndrs, line 29: could this be a simpler test value? 😝

src/cookies.test.ts
```diff
@@ -0,0 +1,67 @@
...
+		expect(getCookie('s_vi')).toEqual(
+			'[CS]v1|29AB343C05013259-4000014980021C3D[CE]',
+		);
+	});
+
+	it('should be able to get country cookie', () => {
```
sndrs, line 35: is this testing anything that the example above isn't testing?

src/cookies.test.ts
```diff
@@ -0,0 +1,67 @@
...
+		document.cookie = 'GU_geo_country=GB;';
+		expect(getCookie('GU_geo_country')).toEqual('GB');
+	});
+
+	it('should be able to set a session cookie', () => {
+		setSessionCookie('cookie-1-name', 'cookie-1-value');
+		expect(document.cookie).toEqual(
```
sndrs, line 42: might be useful to `expect` this to be null before setting it?

src/cookies.test.ts
```diff
@@ -0,0 +1,67 @@
...
+			'cookie-1-name=cookie-1-value; path=/; domain=.theguardian.com',
+		);
+	});
+
+	it('should be able the clean a list of cookies', () => {
```
sndrs, line 47:
```suggestion
	it('should be able to clear a list of cookies', () => {
```

src/cookies.ts
```diff
@@ -0,0 +1,73 @@
...
+export const getCookie = (name: string): string | null => {
+	const cookieVal = getCookieValues(name);
+
+	if (cookieVal.length > 0) {
+		return cookieVal[0];
+	}
+	return null;
+};
+
+export const cleanUp = (names: string[]): void => {
```
sndrs, line 69: how do you feel about exporting `removeCookie` and removing this method? users can map over a list of cookies if they need, and the export name is then super clear about what is happening?

src/cookies.ts
```diff
@@ -0,0 +1,73 @@
...
+
+	// Remove cookie, implicitly using the document's domain.
+	document.cookie = `${name}=;${path}${expires}`;
+	if (!currentDomainOnly) {
+		// also remove from the short domain
+		document.cookie = `${name}=;${path}${expires} domain=${getShortDomain()};`;
+	}
+};
+
+export const setSessionCookie = (name: string, value: string): void => {
```
sndrs, line 53: for completeness, do you want to add a `setCookie` method?

src/cookies.ts
```diff
@@ -0,0 +1,73 @@
...
+	if (!isValidCookieValue(name) || !isValidCookieValue(value)) {
+		return;
+	}
+	document.cookie = `${name}=${value}; path=/;${getDomainAttribute()}`;
+};
+
+export const getCookie = (name: string): string | null => {
+	const cookieVal = getCookieValues(name);
```
sndrs, line 61: getting cookies is a blocking action – is this something we could safely memoize do you think?

src/getLocale.ts
```diff
@@ -1,17 +1,16 @@
+import { getCookie, setSessionCookie } from './cookies';
 import { isString } from './isString';
 import { storage } from './storage';
 import type { CountryCode } from './types/countries';
 
-const KEY = 'gu.geolocation';
+const KEY = 'GU_geo_country';
+const KEY_OVERRIDE = 'gu.geo.override';
```
sndrs, line 7: 🤩

src/getLocale.ts
```diff
@@ -1,17 +1,16 @@
...
 const URL = 'https://api.nextgen.guardianapps.co.uk/geolocation';
 
 // best guess that we have a valid code, without actually shipping the entire list
 const isValidCountryCode = (country: unknown) =>
 	isString(country) && /^[A-Z]{2}$/.test(country as string);
```
sndrs, line 12: i know it's not part of this PR but it would be good to cache this regex rather than recreating it every time this is called

src/index.ts
```diff
@@ -6,4 +6,5 @@ export { isUndefined } from './isUndefined';
 export { loadScript } from './loadScript';
 export { storage } from './storage';
 export { timeAgo } from './timeAgo';
+export { getCookie } from './cookies';
```
sndrs, line 9: i think this is missing some exports?

src/getLocale.ts
```diff
@@ -25,10 +24,19 @@ export const __resetCachedValue = (): void => (locale = void 0);
 export const getLocale = async (): Promise<CountryCode | null> => {
 	if (locale) return locale;
 
-	const stored = storage.local.get(KEY) as CountryCode;
+	// return overrode geo from localStorage, used for changing geo only for development purposes
```
sndrs, line 27:
```suggestion
	// return overridden geo from localStorage, used for changing geo only for development purposes
```

sndrs (changes requested):
this looks really good, so much better than hitting an endpoint! 

couple of things to look at and a couple of questions too

src/cookies.README.md
ioanna0: Yes totally wrong! 😳 fixed now

src/cookies.test.ts
ioanna0: lol indeed, I got it from `frontend` I think a much simpler cookie value is also testing enough of the logic changed it to a trimmed value

src/cookies.test.ts
ioanna0: Nope! the more advanced cookie above is covering what we need.

src/cookies.test.ts
ioanna0: this now tests the `remove` method instead

src/index.ts
ioanna0: Exported all of them now! Ready to be used as a replacement module of [frontend cookies.js](https://github.com/guardian/frontend/blob/main/static/src/javascripts/lib/cookies.js)

src/cookies.ts
ioanna0: i think that's a great idea that would be mostly applicable for cookies that don't change their values for the duration of the session. i don't know if we can assume this would be the case for all cookies? 🤔
maybe we can do it for only specific cookies that we def know they won't change such as country code?

mxdvl: Maybe adding an extra optional parameter like `memoizeValue = true` would be enough? Is it worth adding JSDoc comments to the lib to describe this behaviour?

```js
/** Reads a cookie, and optionally memoizes its value.
 * @param {string} name - the cookie’s name.
 * @param {boolean} memoizeValue - if set to `false`, calls will not memoize the cookie’s value.
 */
export const getCookie = (name: string, memoizeValue = true): string | null => {
  /* … */
}

```

ioanna0: Good idea, I added the boolean in the `setCookie` and `setSessionCookie` instead

src/getLocale.ts
ioanna0: which one is the expensive part? should we memoize the results of calling `test`, or do you mean just the pattern bit?

src/cookies.ts
```diff
@@ -0,0 +1,129 @@
...
+ * @param {number} daysToLive - expiry date will be calculated mased on the daysToLive
+ * @param {boolean} isCrossSubdomain - specify if it's a cross subdomain cookie, default false
+ */
+export const setCookie = (
+	name: string,
+	value: string,
+	daysToLive?: number,
+	isCrossSubdomain = false,
+	shouldMemoize = false,
+): void => {
```
sndrs, line 57: have you considered using a params object here? it would enable to you write (and later read) uses like this:

```js
setCookie({ name: "a", value: "b", isCrossSubdomain: true })
setCookie({ name: "a", value: "b", daysToLive: 1, shouldMemoize: true })
```

here's a [example in typescript](https://www.typescriptlang.org/play?#code/MYewdgzgLgBBCmUDCIQGsCW8YF4YAoBvAKBhjAEMBbeAGlJgDcKAbAVzoYBMKBPCACogAMhkacyGCEgBOICBADKbAEZcQVChjC4YAM1YJ6ZCAAsQbFlwCy8KiAwAvbHgMsjxAL4AuGCTKUNL7QMtoA5gDcDMzs8MFQoWCR3HyCImLwAPy+YGxUKvAyUZLScgrKahpaYNkwKqgs8BRgxXDmljZ2Ds619SCNzVGeAJS4AHx+DKCQ-fAAdCwgYUTk1HRMrBy0MDz8QqLi21Ky8kqq6pra22YWVrb2TtgjQ1HECMiomPArgXEwAEQUf4wZ5vRAodBYH5rXyA-7bGIcWEqYGg94Qr7QoIAoEIzZ-f4o7a7NIHP4ARhBwyi6M+UMIq2xcLxsWR8J2qX2GV85KOpVOFQu1V8CQ4VJp4Lp3wZv1huI2rIBRJgxzKZ0qlzAIpkYtBQA)

src/cookies.ts
```diff
@@ -0,0 +1,129 @@
...
+		},
+	)}`;
+
+	if (shouldMemoize) {
+		memoizedCookies.set(name, getCookieValues(name)[0]);
+	}
+};
+
+/**
+ * Set a session cookieookie. If it's been memoized it won't retrieve it again.
```
sndrs, line 83:
```suggestion
 * Set a session cookieookieookie. If it's been memoized it won't retrieve it again.
```

src/cookies.test.ts
```diff
@@ -0,0 +1,96 @@
...
+		cookieValue = '';
+	});
+
+	it('should be able to get a cookie', () => {
+		document.cookie =
+			'optimizelyEndUserId=oeu1398171767331r0.5280374749563634; __qca=P0-938012256-1398171768649;';
+		expect(getCookie('__qca')).toEqual('P0-938012256-1398171768649');
+	});
+
+	it('should be able to get a memoized cookie', () => {
```
sndrs, line 47: i think not all the ways this could be called are captured here? e.g. 

```js
setCookie('a', 'b', , true)
setCookie('a', 'b', 1, , true)
```

sndrs (changes requested):
couple of test cases missing, also wondered what you thought about using a params rather than a list of args?

src/getLocale.ts
sndrs: yeah that's one, saves recreating a regex for `/^[A-Z]{2}$/` each time it runs – thanks :)

src/cookies.ts
ioanna0: I just changed it again sorry! i think what Max suggested initially to memoize on `getCookie` might be better because we want to memoize on fetch time. We cannot assume that a setCookie has happened in the same session as a get i think? And I changed the setCookie to override the memoized value if it exists.
But great idea with params object, could be useful if we are back to 5 params again

ioanna0: https://github.com/guardian/libs/pull/162/commits/ef4c8daa1279f93ef6886209d649a12f9d307f33

sndrs: i wonder if memoizing cookie reading is really just a premature optimisation that only adds complexity? (sorry i know i brought it up!)

src/cookies.README.md
```diff
@@ -0,0 +1,132 @@
...
+    getCookie,
+    removeCookie,
+    setCookie,
+    setSessionCookie
+ } from '@guardian/libs';
+```
+
+## Methods
+
+-   [`setCookie(name, value, daysToLive, isCrossSubdomain, shouldMemoize)`](#setCookie)
```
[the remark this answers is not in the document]
ioanna0: changed to: `setCookie({name, value, daysToLive?, isCrossSubdomain?})``

src/cookies.ts
ioanna0: maybe but it could also be really useful! hopefully it will be used by consumers. since it's only optional functionality maybe there's no harm leaving it in? it can also be refactored if needs be

src/cookies.ts
```diff
@@ -0,0 +1,137 @@
...
+	const shortDomain = getShortDomain({ isCrossSubdomain });
+	return shortDomain === 'localhost' ? '' : ` domain=${shortDomain};`;
+};
+
+/**
+ * Set a cookie. If it's been memoized it will replace it's memoized value
+ * @param {string} name - the cookie’s name.
+ * @param {string} value - the cookie’s value.
+ * @param {number} daysToLive - expiry date will be calculated mased on the daysToLive
+ * @param {boolean} isCrossSubdomain - specify if it's a cross subdomain cookie, default false
```
sndrs, line 48: i think these are out of date now?

src/cookies.README.md
```diff
@@ -0,0 +1,126 @@
...
+    getCookie,
+    removeCookie,
+    setCookie,
+    setSessionCookie
+ } from '@guardian/libs';
+```
+
+## Methods
+
+-   [`setCookie({name, value, daysToLive, isCrossSubdomain})`](#setCookie)
```
sndrs, line 18:
```suggestion
-   [`setCookie({name, value, daysToLive?, isCrossSubdomain?})`](#setCookie)
```

src/cookies.ts
```diff
@@ -0,0 +1,137 @@
...
+		memoizedCookies.set(name, getCookieValues(name)[0]);
+	}
+};
+
+/**
+ * Set a session cookie. If it's been memoized it will replace memoized value
+ * @param {string} name - the cookie’s name.
+ * @param {string} value - the cookie’s value.
+ */
+export const setSessionCookie = (name: string, value: string): void => {
```
sndrs, line 91: sorry to be a pain, but do you think it would be clearer if all cookie methods took a consisten type of payload – i.e. a params object like `setCookie`?

sndrs: i feel like i would expect that having done 
```js
setCookie({ name: 'a', value: 'b' })
``` 
i'd expect to be able to do something like this too: 
```js
setSessionCookie({ name: 'a', value: 'b' })
getCookie({ name: 'a', shouldMemoize: false })
```

sndrs (changes requested):
some docs outdated now i think, and a question around signature consistency

src/cookies.ts
ioanna0: Yes, though so but at that time i though it was simpler to pass params for 1 or 2 values rather than config obj. Is better to be consistent though across the whole module and have refactored it now across all those methods

src/cookies.ts
ioanna0: how does that look like?
```
 * @param details Details about the cookie.
 * @param details.name - the cookie’s name.
 * @param details.value - the cookie’s value.
 * @param details.daysToLive - optional expiry date will be calculated based on the daysToLive
 * @param details.isCrossSubdomain - specify if it's a cross subdomain cookie, default false
```

src/cookies.ts
```diff
@@ -0,0 +1,149 @@
...
+	}
+	document.cookie = `${name}=${value}; path=/;${getDomainAttribute()}`;
+
+	// If the cookie is already memoized we want to replace its value
+	if (memoizedCookies.has(name)) {
+		memoizedCookies.set(name, getCookieValues(name)[0]);
+	}
+};
+
+export const removeCookie = (name: string, currentDomainOnly = false): void => {
```
sndrs, line 111: this is the only one now not accepting an object – sorry, can it?

sndrs: also, i don't think `currentDomainOnly` is documented?

sndrs (changes requested):
nearly – just these last 2 things then all good

src/cookies.test.ts
```diff
@@ -0,0 +1,172 @@
...
+			cookies.getCookie({ name: 'GU_geo_country', shouldMemoize: true }),
+		).toEqual('IT');
+		expect(
+			cookies.getCookie({ name: 'GU_geo_country', shouldMemoize: true }),
+		).toEqual('IT');
+		expect(
+			cookies.getCookie({ name: 'GU_geo_country', shouldMemoize: true }),
+		).toEqual('IT');
+		// for some reason the spy is been called 1 additional time although that's not happening in reality
+		expect(spy).not.toHaveBeenCalledTimes(2);
```
mxdvl, line 124: Don't you want to set a specific number of times it's expected to be called?

```suggestion
		expect(spy).toHaveBeenCalledTimes(1);
```

src/cookies.ts
```diff
@@ -0,0 +1,161 @@
...
+			);
+		}
+
+		return acc;
+	}, []);
+};
+
+// subset of https://github.com/guzzle/guzzle/pull/1131
+const isValidCookieValue = (name: string) =>
+	!/[()<>@,;"\\/[\]?={} \t]/g.test(name);
```
mxdvl, line 21: Is it worth extracting this Regex to a constant, as we've done for the `validCountryCode`?

sndrs (dismissed):
great! thanks so much for your patience @ioanna0. we discussed this offline, but with a lib like this the API gets set so early on that (hopefully) there's a lot of value in persevering at this stage, so thank you for bearing with all the change requests :)

excellent addition 🎉

src/cookies.ts
sndrs: good spot

src/cookies.test.ts
ioanna0: thanks, good catch! I completely missed the `not` and now is making total sense, is calling it just once ! 🙂

mxdvl: Yes, exactly, It was **not** _not_ doing what you wanted! 😃

sndrs (approved):
lovely stuff :)

sndrs:
AOK to merge?

ioanna0:
Yes!

PR #48, opened 2020-11-06
feat: add getSwitches

sndrs:
##  What does this change?

adds the following:

- `getSwitches`

gets value from `window.guardian.config` if present, or from a new endpoint (`~~coming from~~ provided by @shtukas ❤️ ) if not.

also adds a couple of helpers:

- `isObject` and `isBoolean`

## Why?

- it will be super useful to be able to use these from other `@guardian/` packages
- https://github.com/orgs/guardian/teams/client-side-infra/discussions/10

PR #163, opened 2021-04-21
feat: `isObject` Function

JamieB-gu:
## What does this change?

- Adds an `isObject` function

## Why?

Checks whether an `unknown` value is an object. Particularly useful when dealing with `JSON.parse`, as seen [here](https://github.com/guardian/apps-rendering/blob/c3cb9c7a25c0124c342408d790ba0e74ba8abcdb/src/client/interactives.ts#L7). Migrating upstream to `libs` from the implementation [in apps-rendering](https://github.com/guardian/apps-rendering/blob/1ed8a5d6208cc3972db05bbb474cd19b77e5ea18/src/lib.ts#L58).

### DIFF
--- a/src/index.ts
+++ b/src/index.ts
@@ -1,10 +1,10 @@
 /* istanbul ignore file */
 
 export { getLocale } from './getLocale';
+export { isObject } from './isObject';
 export { isString } from './isString';
 export { isUndefined } from './isUndefined';
 export { loadScript } from './loadScript';
 export { storage } from './storage';
 export { timeAgo } from './timeAgo';
 export { log, debug } from './logger';
-export { isObject } from './isObject';
--- a/src/index.ts
+++ b/src/index.ts
@@ -7,3 +7,4 @@ export { loadScript } from './loadScript';
 export { storage } from './storage';
 export { timeAgo } from './timeAgo';
 export { log, debug } from './logger';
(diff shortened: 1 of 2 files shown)


sndrs (changes requested):
cheers for this. 

there's actually [an implementation of this](https://github.com/guardian/libs/pull/48/files#diff-f560ab256615b1dd535259c3d96a113d4620cd13cec5d308293e710e65535234) floating around in #48 that i need to merge!

in that instance I just cribbed sindre sorhus's [is-plain-obj](https://github.com/sindresorhus/is-plain-obj) method. any thoughts on the implementations?

also, there's a test in #48 that would be good to copy over to this PR

JamieB-gu:
> any thoughts on the implementations?

Don't feel strongly on the implementation, whichever is more accurate works for me 👍.

I'm interested in the return type though. Strictly speaking isn't `Record<string | number, unknown>` essentially equivalent to `Record<string, unknown>`?

```ts
const a = { 4: 5 };
const b = { '4': 5 };

console.log(a[4]); // Prints 5
console.log(a['4']); // Prints 5
console.log(b[4]); // Prints 5
console.log(b['4']); // Prints 5
```

sndrs:
closing as #48 added `isObject`

PR #171, opened 2021-05-07
feat: Parsing Lib

JamieB-gu:
## What does this change?

This adds a library for parsing `unknown` values into other TypeScript types. It's generic enough to work for any `unknown` value, but is probably most useful for parsing JSON.

It's inspired by the [Elm JSON module](https://package.elm-lang.org/packages/elm/json/latest/), Haskell modules like [ReadP](https://hackage.haskell.org/package/base-4.14.1.0/docs/Text-ParserCombinators-ReadP.html), and various other parser-combinator libraries. It also uses the `Result` and `Option` types available in our [`@guardian/types`](https://github.com/guardian/types) library.

## Why?

[`JSON.parse`](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/JSON/parse) and related methods like [`Body.json`](https://developer.mozilla.org/en-US/docs/Web/API/Body/json) from the Fetch API return their parsed JS data structures as type [`any`](https://www.typescriptlang.org/docs/handbook/2/everyday-types.html#any). The use of `any` is a bit like an escape hatch from the TypeScript type system, and can result in runtime errors elsewhere in the codebase as the value with this type gets passed around and used.

One way around this is to type the returned value from `JSON.parse` as [`unknown`](https://www.typescriptlang.org/docs/handbook/2/functions.html#unknown) (as discussed in [this issue](https://github.com/typescript-eslint/typescript-eslint/issues/2118#issuecomment-641464651)). TypeScript will then _force_ you to verify the structure of this value, through methods like [type guards](https://www.typescriptlang.org/docs/handbook/2/narrowing.html#typeof-type-guards) and [type predicates](https://www.typescriptlang.org/docs/handbook/2/narrowing.html#using-type-predicates), before using it. This should significantly reduce the possibility of runtime errors.

A downside to taking this approach, however, is that it can result in quite verbose, repetitive code - particularly when working with complex, nested objects. This is where this library comes in. It attempts to simplify the parsing process by abstracting the repetitive manual checks behind a declarative-style API, based on parser combinators.

## API

I've documented the API inline using JSDocs; here's a facsimile of those docs for convenience.

### `succeed: <A>(a: A) => Parser<A>`

Ignores whatever the input is and instead provides a successful parsing result containing `a`.

<dl>
  <dt>Parameter <code>a</code></dt>
  <dd>The value to be contained in the successful parsing result</dd>
  <dt>Returns</dt>
  <dd>A <code>Parser</code> of <code>A</code></dd>
</dl>

#### Example

```ts
const parseFoo = succeed('foo'); // Parser<string>

const result = run(parseFoo)(42); // Ok<string>, with value 'foo'
```

### `fail: <A>(e: string) => Parser<A>`

Ignores whatever the input is and instead provides a failed parsing result containing the error string `e`.

<dl>
  <dt>Parameter <code>e</code></dt>
  <dd>The error string to be provided as the parsing error</dd>
  <dt>Returns</dt>
  <dd>A <code>Parser</code> of <code>A</code></dd>
</dl>

#### Example

```ts
const parseFoo = fail('Uh oh!'); // Parser<A>

const result = run(parseFoo)(42); // Err<string>, with value 'Uh oh!'
```

### `run: <A>(pa: Parser<A>) => (a: unknown): Result<string, A>`

Run a parser over a given input that's currently `unknown`, and attempt to parse that input into a value of type `A`. If the parsing was successful the result will be an `Ok<A>`. If it failed the result will be an `Err<string>`.

<dl>
  <dt>Parameter <code>pa</code></dt>
  <dd>The parser to run over the input</dd>
  <dt>Parameter <code>a</code></dt>
  <dd>The <code>unknown</code> input to be parsed into type-safe value of type <code>A</code></dd>
  <dt>Returns</dt>
  <dd>A parsing result, wrapped in a <code>Result</code> type</dd>
</dl>

#### Example

```ts
const json: unknown = JSON.parse('{ "foo": 42 }');

const parserA = parseField('foo', parseNumber); // Parser<number>
const resultA = run(parserA)(json); // Ok<number>, with value 42

const parserB = parseField('bar', parseNumber); // Parser<number>
const resultB = run(parserB)(json); // Err<string>, with 'missing field' err
```

### `parseString: Parser<string>`

Parses a value into a `string`.

### `parseNumber: Parser<number>`

Parses a value into a `number`.

### `parseDate: Parser<Date>`

Parses a value into a `Date`.

### `maybe: <A>(pa: Parser<A>) => Parser<Option<A>>`

Makes the value handled by the given parser optional.

**Note:** This will effectively absorb any failure to parse the value, converting it to a `None` instead.

<dl>
  <dt>Parameter <code>pa</code></dt>
  <dd>A parser</dd>
  <dt>Returns</dt>
  <dd>A new <code>Parser</code> with an optional value</dd>
</dl>

#### Example

```ts
const json: unknown = JSON.parse('{ "foo": 42 }');

const parserA = maybe(parseField('bar', parseNumber)); // Parser<Option<number>>
const resultA = run(parserA)(json); // Ok<None>

const parserB = parseField('bar', maybe(parseNumber)); // Parser<Option<number>>
const resultB = run(parserB)(json); // Err<string>, with 'missing field' err
```

### `parseField: <A>(field: string, pa: Parser<A>) => Parser<A>`

Runs the parser `pa` over the value found at a given field in an object. Will fail if the input isn't an object or the field is missing.

<dl>
  <dt>Parameter <code>field</code></dt>
  <dd>The field containing the value to parse</dd>
  <dt>Parameter <code>pa</code></dt>
  <dd>The parser to use on the field value</dd>
  <dt>Return</dt>
  <dd>A parser for the value at the given field</dd>
</dl>

#### Example

```ts
const json: unknown = JSON.parse('{ "foo": 42 }');

const parserA = parseField('foo', parseNumber); // Parser<number>
const resultA = run(parserA)(json); // Ok<number>, with value 42

const parserB = parseField('bar', parseNumber); // Parser<number>
const resultB = run(parserB)(json); // Err<string>, with 'missing field' err
```

### `parseIndex: <A>(index: number, pa: Parser<A>) => Parser<A>`

Runs the parser `pa` over the value found at a given index in an array. Will fail if the input isn't an array or the index is empty.

<dl>
  <dt>Parameter <code>index</code></dt>
  <dd>The array index containing the value to parse</dd>
  <dt>Parameter <code>pa</code></dt>
  <dd>The parser to use on the value at the given index</dd>
  <dt>Return</dt>
  <dd>A parser for the value at the given index</dd>
</dl>

#### Example

```ts
const json: unknown = JSON.parse('[41, 42, 43]');

const parserA = parseIndex(1, parseNumber); // Parser<number>
const resultA = run(parserA)(json); // Ok<number>, with value 42

const parserB = parseIndex(7, parseNumber); // Parser<number>
const resultB = run(parserB)(json); // Err<string>, with 'missing index' err
```

### `parseAt: <A>(location: string[], pa: Parser<A>) => Parser<A>`

A convenience for parsing values in nested objects. Takes a list of fields used to pinpoint a location within the nested objects.

<dl>
  <dt>Parameter <code>location</code></dt>
  <dd>Field names designating a location inside a nested object</dd>
  <dt>Parameter <code>pa</code></dt>
  <dd>The parser to use on the value at the location</dd>
  <dt>Return</dt>
  <dd>A parser for the value at the given location</dd>
</dl>

#### Example

```ts
const json: unknown = JSON.parse('{ "foo": { "bar": 42 } }');

const parser = parseAt(['foo', 'bar'], parseNumber); // Parser<number>
const result = run(parser)(json); // Ok<number>, with value 42
```

### `parseArray: <A>(pa: Parser<A>) => Parser<A[]>`

Parses an array containing values of type `A`. Will fail if the input isn't an array, or the values can't be parsed as `A`.

<dl>
  <dt>Parameter <code>pa</code></dt>
  <dd>A parser for the values in the array</dd>
  <dt>Return</dt>
  <dd>A parser for the array</dd>
</dl>

#### Example

```ts
const json: unknown = JSON.parse('[41, 42, 43]');

const parser = parseArray(parseNumber); // Parser<number[]>
const result = run(parser)(json); // Ok<number[]>, with value [41, 42, 43]
```

### `map: <A, B>(f: (a: A) => B) => (pa: Parser<A>) => Parser<B>`

Converts a `Parser` of `A` to a `Parser` of `B`. Will apply the given function `f` to the result of the given parser (`Parser<A>`) if that parser is successful.

<dl>
  <dt>Parameter <code>f</code></dt>
  <dd>The function to apply</dd>
  <dt>Parameter <code>pa</code></dt>
  <dd>The parser to convert</dd>
  <dt>Return</dt>
  <dd>A new parser for type <code>B</code></dd>
</dl>

#### Example

```ts
const parser = map(n => n + 1)(parseNumber); // Parser<number>

const valueA: unknown = 41;
const resultA = run(parser)(valueA); // Ok<number>, with value 42

const valueB: unknown = 'foo';
const resultB = run(parser)(valueB); // Err<string>, with 'not a number' err
```

### `map2: <A, B, C>(f: (a: A, b: B) => C) => (pa: Parser<A>, pb: Parser<B>) => Parser<C>`

Similar to `map`. Will apply the given function `f` to the results of two given parsers (`Parser<A>` and `Parser<B>`) if both of those parsers are successful.

<dl>
  <dt>Parameter <code>f</code></dt>
  <dd>A function with two arguments, <code>a</code> and <code>b</code>, which correspond to the values potentially returned by the two parsers</dd>
  <dt>Parameter <code>pa</code></dt>
  <dd>The first parser, for a value of type <code>A</code></dd>
  <dt>Parameter <code>pb</code></dt>
  <dd>The second parser, for a value of type <code>B</code></dd>
  <dt>Return</dt>
  <dd>A new parser for type <code>C</code></dd>
</dl>

#### Example

```ts
type Person = { name: string, age: number };

const makePerson = (name: string, age: number): Person => ({ name, age });
const parsePerson: Parser<Person> =
  map2(makePerson)(
    parseField('name', parseString),
    parseField('age', parseNumber),
  );

const jsonA: unknown = JSON.parse('{ "name": "CP Scott", "age": 85 }');
const resultA = run(parsePerson)(jsonA); // Ok<Person>

const jsonB: unknown = JSON.parse('{ "name": "CP Scott" }');
const resultB = run(parsePerson)(jsonB); // Err<string>, 'missing field' err
```

### `map3`, `map4`, `map5`, `map6`, `map7`, `map8`

All similar to `map2`, but for more parsers. See the docs for that function for more details and examples.

### `andThen: <A, B>(f: (a: A) => Parser<B>) => (pa: Parser<A>) => Parser<B>`

Like `map` but applies a function that *also* returns an `Parser`. Then "unwraps" the result for you so you don't end up with `Parser<Parser<A>>`. Can be useful for stringing together multiple parsing steps, where each step relies on the result of the previous one.

If the first parser fails, the function won't be called and the error will be returned instead. If the second parser fails the error for that will be returned.

<dl>
  <dt>Parameter <code>f</code></dt>
  <dd>The function to apply</dd>
  <dt>Parameter <code>pa</code></dt>
  <dd>The parser whose result is to be passed to the function</dd>
  <dt>Return</dt>
  <dd>A new parser, built from <code>pa</code> and <code>f</code></dd>
</dl>

#### Example

```ts
type MultipleChoiceAnswer = 'a' | 'b' | 'c';

const parseAnswerValue = (value: string): Parser<MultipleChoiceAnswer> =>
  value === 'a' || value === 'b' || value === 'c'
    ? succeed(value)
    : fail("Needed 'a', 'b' or 'c'.");

const parseAnswer = pipe(
  parseField('answer', parseString),
  andThen(parseAnswerValue),
);

const jsonA: unknown = JSON.parse('{ "answer": "a" }');
const resultA = run(parseAnswer)(jsonA); // Ok<MultipleChoiceAnswer>, 'a'

const jsonB: unknown = JSON.parse('{ "answer": "d" }');
const resultB = run(parseAnswer)(jsonB); // Err<string>, "Needed 'a', 'b' or 'c'."
```

### `oneOf: <A>(pas: Array<Parser<A>>) => Parser<A>`

Docs todo.


sndrs:
💥 wow this is a biggie...

i'm not sure how to proceed with reviewing this, because i'm not 100% clear on the motivation and what's going on if i'm honest...

here's some initial thoughts:

1. would you be able to provide an example of the kind of code this is trying to improve and a fuller demonstration of how this helps?
2. should this live in here? libs is "A collection of JavaScript libraries for Guardian projects" – does this help people who aren't using TS?
3. this repo currently uses a convention of providing a README alongside the code by way of docs. it should probably move to auto-generating docs from jsdoc comments but in lieu of that it would be good to be consistent and provide the same way of understanding how each module works and what it offers.

JamieB-gu:
> here's some initial thoughts:

Thanks for taking a look @sndrs! I've made some changes to address your comments:

> 1. would you be able to provide an example of the kind of code this is trying to improve and a fuller demonstration of how this helps?

I've added an example to the new `README` which I hope will demonstrate usage.

As to the problems this solves, an example can be found described in this PR: https://github.com/guardian/apps-rendering/pull/1298. We were receiving some JSON from one of MAPI's sport endpoints, and it was possible for that JSON to vary in structure. Attempting to use type casting resulted in runtime crashes, and error messages that were difficult to trace back to the original problem. Use of this library will prevent those crashes, and provide better error messages for debugging.

There are several other examples of runtime crashes caused by unexpected JSON across different projects in the department. Happy to chat through some of them if you'd like?

> 2. should this live in here? libs is "A collection of JavaScript libraries for Guardian projects" – does this help people who aren't using TS?

Yes, this will also work in JS-land. I've added a note to the new `README` to clarify this.

> 3. this repo currently uses a convention of providing a README alongside the code by way of docs. it should probably move to auto-generating docs from jsdoc comments but in lieu of that it would be good to be consistent and provide the same way of understanding how each module works and what it offers.

Good point! Added 📜.

JamieB-gu:
@sndrs @SiAdcock I'm partway through adding tests. Hopefully the ones I've added so far can help with understanding how this module works?

sndrs:
**_tl;dr_**

I'm afraid i'm really struggling with the API and i don't think I can't usefully review it. Would you consider rewriting it using  native J/TS idioms?

**_long version_**

I've tried to review this a few times now, but I can't wrap my head around the API 😕. I'm not familiar enough with the languages that inspired the patterns it uses, which means I don't think I can really review it as it stands. 

But, even if someone else more familiar with the patterns in the code is happy with it, I still do not actually think it should be merged in its current state either, I'm afraid. 

And here's why: 

things that live in `libs` should be as straightforward and easy to understand as possible – it's intended to quickly and clearly capture best practices, save you time and allow you to focus on what you're doing.

This PR adopts patterns that are pretty alien to (at least my experience of) typescript. 

I don't think that this module, as this stands, will provide that quick import-and-away-you-go solution for anyone who (like me) is not familiar with elm or Haskell. 

It can't be the only way of doing it, and i think it's hard to justify needing to know the idioms of an unrelated language to use a standard javascript/typescript library at the guardian.

So on that basis alone, I don't think it's appropriate for a universal toolkit like `libs`.

_Whether or not it _should_ be appropriate is beyond the bounds of this review, but by way of example, I don't think any codebase outside apps rendering (or its offspring) [uses `Option`/`Result` from `types`](https://github.com/search?p=1&q=org%3Aguardian+%22guardian%2Ftypes%22+Option&type=Code). For good or bad, this way of writing J/TS is not common at the guardian._

That is totally not to say we shouldn't have a codified way of safely working with json, or that `libs` is the wrong place for it. 

But _that_ then raises a 2nd problem: 

because I'm struggling to understand the API, I cannot actually say if i think your module is something we should adopt, or whether we could save some maintenance cost, and recommend (or wrap) something from https://json-schema.org/implementations.html#validator-javascript, for instance. 

And this is a real shame, because the guardian is potentially losing out on something valuable not because of what it does or doesn't do, but because the way it's written effectively makes it inaccessible to anyone like me who knows the language it's written in, but not the languages it's written _like_.

You've _clearly_ put a lot of work into this already, but would you consider trying to see if it's possible to accomplish the same thing, but using more J/TS-like patterns? 🙏

(I'd also be very interested to hear other people's takes on this...)

JamieB-gu:
> I've tried to review this a few times now, but I can't wrap my head around the API 😕. I'm not familiar enough with the languages that inspired the patterns it uses, which means I don't think I can really review it as it stands... I still do not actually think it should be merged in its current state either, I'm afraid.

No worries! Thanks for giving it a fair shake. I had a conversation with @SiAdcock the other day that went along similar lines.

> would you consider investigating whether it's possible to accomplish the same thing, but using more J/TS-like patterns? 🙏

I can try 🤔 - there are a couple of hurdles I can think of:

1. Unfortunately, this library is built on a "parser combinator" pattern that's fairly common in other libraries - I wasn't so much inventing something new as writing a TS implementation. To build it another way I'd have to come up with something new and at the moment I don't have a good idea for _how_ to do that, but I'll give it some thought...

2. I'd need to understand what "J/TS-like patterns" are from your perspective. To me this isn't a particular unusual pattern for TypeScript (although perhaps less common in JS-land?). I guess it uses a fairly functional style and folks may be more used to procedural/imperative-style code instead? Perhaps if you could give me an idea of the features that you find problematic it would help guide me towards an implementation that's more palatable? For example, if you could tell me which of the following you think are reasonable vs those you think would cause confusion?

- Generics
- Curried/partial-applied functions
- De-structuring assignment
- Spread/rest parameters
- Recursion
- `Option`/`Result` types
- `map` and `andThen`
- JSDocs
- The `Parser` type
- Higher-order functions

> I don't think any codebase outside apps rendering (or its offspring) uses `Option`/`Result` from `types`

Fair point. Simon also mentioned that it's difficult to judge how it would be used based on the minimal examples included. Perhaps it would be sensible to incubate this within the Apps team for now. That would be helpful because a) we have a couple of pressing use cases that we'd like to use it for, b) we can determine whether it's useful in production, or whether folks find it confusing, and c) it would provide those examples of real-world use.

(**Note:** `mobile-purchases` actually includes a [variant of `Option`](https://github.com/guardian/mobile-purchases/blob/fb8d6d6d27e45859b86af807246786837f407497/typescript/src/utils/option.ts), and it wasn't even me who added it 🙂)

> I'm afraid i'm really struggling with the API and i don't think I can't usefully review it.

I'm happy to sit down (virtually) and walk through the code with you if you'd like? That won't solve the concerns you have about what's appropriate for this codebase/the department, but it might be interesting nonetheless?

> That is totally not to say we shouldn't have a codified way of safely working with json

> I cannot actually say if i think your module is something we should adopt, or whether we could save some maintenance cost, and recommend (or wrap) something...

Yeah it's definitely a problem. It was originally someone posting in the CS channel about the tribulations of working with un-typed JSON that gave me the idea (I can't remember who). That, coupled with the use-cases we have in Apps that I mentioned above, suggest to me that it's not an uncommon issue within the dept 😬.

sndrs:
> Unfortunately, this library is built on a "parser combinator" pattern that's fairly common in other libraries - I wasn't so much inventing something new as writing a TS implementation. To build it another way I'd have to come up with something new and at the moment I don't have a good idea for how to do that, but I'll give it some thought...

i'm not 100% what the library is trying achieve, but is it to avoid writing something like this?

```ts
type EndpointResponse = {
  a?: {
    b?: {
      c?: string | number,
    }
  }
}

const json = { a: { b: { c: 'hi' } } } as EndpointResponse;

const c = json.a?.b?.c

switch (typeof c) {
  case 'string':
    doSomethingWithAString(c) 
    break;
  case 'number':
    doSomethingWithANumber(c) 
    break;
  default:
    return;
}

```

JamieB-gu:
> i'm not 100% what the library is trying achieve, but is it to avoid writing something like this?

Essentially yes. That sort of type coercion (`as`) is quite brittle, and can result in runtime errors that are difficult to track down because they're thrown much later in the code. Also, the parsing code to mitigate them can get _very_ verbose _very_ quickly, and as a result it can be difficult to ensure you've handled all cases.

[This PR](https://github.com/guardian/mobile-purchases/pull/259/files#diff-4462a45fd373342ca1a270641d6f170070cf6788e1868b15337b1f8d52033f27) was opened to address a bug caused by the first problem, and in fixing it demonstrates the second problem.

PR #181, opened 2021-05-11
refactor!: types are exported from the package root and we use our types library

sndrs:
## What does this change?

#157 changed the shape of `dist`, so that types are now available from the package root rather than alongside JS, so we need to do a major bump cf https://github.com/guardian/dotcom-rendering/blob/main/src/web/browser/lotame/init.ts#L4

this PR also removes the `Countries` and `CountryCodes` types and uses the ones added to `@guardian/types`.


README.md
```diff
@@ -4,7 +4,6 @@
 [![ES version](https://badgen.net/badge/ES/2020/cyan)](https://tc39.es/ecma262/2020/)
 [![npm type definitions](https://img.shields.io/npm/types/@guardian/libs)](https://www.typescriptlang.org/)
 [![Coverage Status](https://coveralls.io/repos/github/guardian/libs/badge.svg)](https://coveralls.io/github/guardian/libs)
-[![gzip size](https://img.badgesize.io/https://unpkg.com/@guardian/libs/dist/umd/index.min.js?compression=gzip)](https://unpkg.com/@guardian/libs/dist/umd/index.min.js)
```
mxdvl, line 7: Why not?

sndrs: Yeah exactly

PR #209, opened 2021-06-11
fix: use setUTC and getUTC methods for setting cookie expiry dates

arelra:
## What does this change?

Update `cookies.ts setCookie()` to set cookie expiry dates with [`setUTC`](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Date/setUTCDate
) and [`getUTC`](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Date/getUTCDate) methods to maintain UTC time.

## Why?

The current logic for setting cookie expiry dates is setting local timezone time rather than maintaining UTC time.

The test for `setCookie` fails because it (correctly) expects the expiry time to be UTC.

Because non-UTC methods are used to set the expiry date, it incorrectly sets the time to local time.

Hence `expires.toUTCString()` outputs local time rather than the expected adjusted UTC time.

### Reproduction

```js
// Date with GMT timezone within BST period
var a = new Date('Mon Jun 14 2021 12:00:00 GMT+0000 (Greenwich Mean Time)');
var b = new Date('Mon Jun 14 2021 12:00:00 GMT+0000 (Greenwich Mean Time)');

// Set date to move into GMT
a.setMonth(10);
b.setUTCMonth(10);

console.log(a.toUTCString()); // Sun, 14 Nov 2021 13:00:00 GMT - incorrect
console.log(b.toUTCString()); // Sun, 14 Nov 2021 12:00:00 GMT - correct
```

### DIFF
--- a/src/cookies.test.ts
+++ b/src/cookies.test.ts
@@ -37,7 +37,7 @@ describe('cookies', () => {
 		MockDate.reset();
 	});
 
-	it('should be able to get a cookie', () => {
+	it('get a cookie', () => {
 		document.cookie =
 			'optimizelyEndUserId=oeu1398171767331r0.5280374749563634; __qca=P0-938012256-1398171768649;';
 		expect(cookies.getCookie({ name: '__qca' })).toEqual(
@@ -45,7 +45,7 @@ describe('cookies', () => {
 		);
 	});
 
-	it('should be able to set a cookie with an expiry date in six months that preserves UTC time', () => {
+	it('set a cookie with an expiry date in six months that preserves UTC time', () => {
 		expect(document.cookie).toEqual('');
 		MockDate.set('Sun Nov 17 2019 12:00:00 GMT+0000 (Greenwich Mean Time)');
 		cookies.setCookie({
@@ -59,7 +59,7 @@ describe('cookies', () => {
 		);
 	});
 
-	it('should be able to set a cookie to expire in a specific number of days', () => {
+	it('set a cookie to expire in a specific number of days', () => {
 		expect(document.cookie).toEqual('');
 		MockDate.set('Sun Nov 17 2019 12:00:00 GMT+0000 (Greenwich Mean Time)');
 		cookies.setCookie({
@@ -72,7 +72,7 @@ describe('cookies', () => {
 		);
 	});
 
-	it('should be able to set a cookie for a specific number of days with an expiry that preserves UTC time', () => {
+	it('set a cookie for a specific number of days with an expiry that preserves UTC time', () => {
 		expect(document.cookie).toEqual('');
 		// BST started Sun 28th Mar 2021
 		MockDate.set('Sat Mar 27 2021 12:00:00 GMT+0000 (Greenwich Mean Time)');
@@ -86,7 +86,7 @@ describe('cookies', () => {
 		);
 	});
 
-	it('should not set a cookie when the cookie name is invalid', () => {
+	it('does not set a cookie when the cookie name is invalid', () => {
 		expect(document.cookie).toEqual('');
 		cookies.setCookie({
 			name: 'cookie-1-name-@',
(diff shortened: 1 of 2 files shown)


sndrs (approved):
this is a great catch!

mxdvl (dismissed):
One of the quirks of living so close to UTC time is you don’t always spot inaccuracies…

sndrs:
i wonder if it's something we should lint against - is there _ever_ a reason to not use a `setUTC` method?

arelra:
> i wonder if it's something we should lint against - is there _ever_ a reason to not use a `setUTC` method?

I don't _think_ so.

In principle we should be using UTC dates everywhere and only use timezones if required for display.

mxdvl (dismissed):
GMT+1

src/cookies.test.ts
```diff
@@ -66,8 +70,36 @@ describe('cookies', () => {
...
+
+	it('does not set a cookie when the cookie name is invalid', () => {
+		cookies.setCookie({
+			name: 'cookie-1-name-@',
+			value: 'cookie-1-value',
+		});
 		expect(document.cookie).toEqual('');
+	});
+
+	it('does not set a cookie when the cookie value is invalid', () => {
```
sndrs, line 94: should it throw in this instance to you let you know?

src/cookies.test.ts
```diff
@@ -66,8 +70,36 @@ describe('cookies', () => {
...
+			name: 'cookie-1-name',
+			value: 'cookie-1-value',
+			daysToLive: 7,
+		});
+		expect(document.cookie).toEqual(
+			'cookie-1-name=cookie-1-value; path=/; expires=Sat, 03 Apr 2021 12:00:00 GMT; domain=.theguardian.com',
+		);
+	});
+
+	it('does not set a cookie when the cookie name is invalid', () => {
```
sndrs, line 86: should it throw in this instance to you let you know?

src/cookies.test.ts
```diff
@@ -66,8 +70,36 @@ describe('cookies', () => {
...
+
+	it('does not set a cookie when the cookie value is invalid', () => {
+		cookies.setCookie({
+			name: 'cookie-1-name',
+			value: 'cookie-1-value-<',
+		});
+		expect(document.cookie).toEqual('');
+	});
+
+	it('set a session cookie', () => {
```
sndrs, line 102: just a typo nbd
```suggestion
	it('sets a session cookie', () => {
```

src/cookies.test.ts
```diff
@@ -77,7 +109,35 @@ describe('cookies', () => {
...
+		});
+		expect(document.cookie).toEqual('');
+		cookies.setSessionCookie({
+			name: 'cookie-1-name',
+			value: 'cookie-1-value',
+		});
+		expect(document.cookie).toEqual('cookie-1-name=cookie-1-value; path=/');
+	});
+
+	it('does not set a session cookie when the cookie name is invalid', () => {
```
sndrs, line 124: could session/non-session tests be merged? not a blocker, no worries if you prefer it this way

sndrs (approved):
a couple of things to address if you think they're valid, but if not then 👍 – great stuff, thanks for picking it up!

src/cookies.test.ts
arelra: Good suggestion. I'll add in another PR so this one remains focussed on the UTC bug fix.

src/cookies.test.ts
arelra: Good suggestion. I'll add in another PR so this one remains focussed on the UTC bug fix.

src/cookies.test.ts
arelra: Good suggestion.

`setSessionCookie` is only used by `libs/getLocale` so we can update this with minimal impact.

 I'll add in another PR so this one remains focussed on the UTC bug fix.

src/cookies.test.ts
arelra: Will pick up in the next PR :)

arelra: I had to rebase so did this anyway with some further tidy up of the test descriptions

sndrs (approved):
lovely stuff :)

PR #212, opened 2021-06-16
feat: migrate countries from types

jamie-lynch:
## What does this change?

* Migrates the Country enum and CountryCode type from `@guardian/types` to `@guardian/libs`
* Renames the `Country` enum to `CountriesByCountryCode`

## Why?


src/index.ts
```diff
@@ -6,6 +6,7 @@ export {
 	setCookie,
 	setSessionCookie,
 } from './cookies';
+export type { Country, CountryCode } from './countries';
```
jamie-lynch, line 9: Should `Country` be exported as a type here?

sndrs: after chatting offline, no :)

src/countries.README.md
```diff
@@ -0,0 +1,9 @@
+# `countries`
```
sndrs, line 1: think this needs updating?

sndrs (changes requested):

src/countries.README.md
jamie-lynch: I've just updated the usage example to `CountriesByCountryCode`. Do you think the filename should be changed too?

sndrs: i've been umming and ahhing about this.

should we expand this to look something like:

```typescript
export const countries = {
	afghanistan: {
		name: 'Afghanistan',
		countryCode: 'AF',
	},
	bolivia: {
		name: 'Bolivia (Plurinational State of)',
		countryCode: 'BO',
	},
} as const;

export type Country = keyof typeof countries;
export type CountryCode = typeof countries[Country]['countryCode'];
export type CountryName = typeof countries[Country]['name'];

let getCountryByCodeCache: {
	[prop in CountryCode]?: CountryName;
} = {};

export const getCountryNameByCountryCode = (countryCode: CountryCode) =>
	(getCountryByCodeCache[countryCode] ||= Object.values(countries).find(
		(country) => country.countryCode === countryCode,
	)?.name);
```

so then

```typescript
import { countries, getCountryNameByCountryCode } from '@guardian/libs';

const afghanNameFromCode = getCountryNameByCountryCode('AF')
const afghanName = countries.afghanistan.name
```

sndrs: (and keep the filename)

sndrs: cc @mxdvl

jamie-lynch: This makes sense to me. I'll have a go at implementing it and update this PR.

mxdvl: Don’t have a strong feeling about this… I guess it’s worth mentioning we already have a mapping between country codes and country names in `frontend` here: https://github.com/guardian/frontend/blob/bb292f016a0089f809dffe2851436efe66cd15fe/static/src/javascripts/lib/geolocation.ts

sndrs (approved):

sndrs (approved):

PR #213, opened 2021-06-16
test: 100% coverage for `timeAgo`

mxdvl:
## What does this change?

Achieve 100% coverage in `timeAgo`.

- remove singular verbose “second ago”
- remove singular “day ago”
- add test for future dates

## Why?

[Greener is better!](https://coveralls.io/builds/40626836/source?filename=src%2FtimeAgo.ts)

### DIFF
--- a/src/timeAgo.test.ts
+++ b/src/timeAgo.test.ts
@@ -220,4 +220,9 @@ describe('timeAgo', () => {
 		expect(timeAgo(eightDaysAgo)).toBe('9 Nov 2019');
 		expect(timeAgo(aWhileBack)).toBe('2 Apr 2017');
 	});
+
+	it('returns false on future dates', () => {
+		const tomorrow = new Date(Date.UTC(2019, 10, 18, 10, 0, 0)).getTime();
+		expect(timeAgo(tomorrow)).toBe(false);
+	});
 });
--- a/src/timeAgo.ts
+++ b/src/timeAgo.ts
@@ -50,8 +50,7 @@ const getSuffix = (type: Unit, value: number, verbose?: boolean): string => {
 	const shouldPluralise = value !== 1;
 	switch (type) {
 		case 's': {
-			if (verbose && shouldPluralise) return ' seconds ago';
-			if (verbose) return ' second ago';
+			if (verbose) return ' seconds ago';
 			return 's ago';
 		}
 		case 'm': {
--- a/src/timeAgo.ts
+++ b/src/timeAgo.ts
@@ -50,6 +50,7 @@ const getSuffix = (type: Unit, value: number, verbose?: boolean): string => {
 	const shouldPluralise = value !== 1;
 	switch (type) {
 		case 's': {
+			// Always pluralised, as less than 15 seconds returns “now”
 			if (verbose) return ' seconds ago';
 			return 's ago';
 		}
@@ -64,6 +65,7 @@ const getSuffix = (type: Unit, value: number, verbose?: boolean): string => {
 			return 'h ago';
 		}
 		case 'd': {
+			// Always pluralised, as less than 2 days returns “Yesterday HH:MM”
 			if (verbose) return ' days ago';
 			return 'd ago';
 		}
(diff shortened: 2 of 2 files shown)


src/timeAgo.test.ts
```diff
@@ -220,4 +220,9 @@ describe('timeAgo', () => {
 		expect(timeAgo(eightDaysAgo)).toBe('9 Nov 2019');
 		expect(timeAgo(aWhileBack)).toBe('2 Apr 2017');
 	});
+
+	it('returns false on future dates', () => {
+		const tomorrow = new Date(Date.UTC(2019, 10, 18, 10, 0, 0)).getTime();
+		expect(timeAgo(tomorrow)).toBe(false);
+	});
```
oliverlloyd, line 227: 👍 nice

mxdvl: `false` 😆

src/timeAgo.ts
```diff
@@ -65,8 +64,7 @@ const getSuffix = (type: Unit, value: number, verbose?: boolean): string => {
 			return 'h ago';
 		}
 		case 'd': {
-			if (verbose && shouldPluralise) return ' days ago';
-			if (verbose) return ' day ago';
+			if (verbose) return ' days ago';
```
oliverlloyd, line 67: Would this not allow the result "Last updated 1 days ago"?

mxdvl: It would return “Last updated yesterday 10:00”. We get hours up to 1 day ago, then “Yesterday HH:MM” up to “two days ago”.

oliverlloyd: Ah ha! Great spot

oliverlloyd (approved):

mxdvl:
Ta, I need a +1 from a `CODEOWNER`, too.

Also: I was investigating this a little more, and wondered what happens during BST…

src/timeAgo.ts
sndrs: do you fancy adding a comment explaining this? and the same with seconds above (<15 = 'now')?

src/timeAgo.test.ts
sndrs: should it return `false` or should it error?

sndrs (changes requested):

oliverlloyd:
> should it return false or should it error?

Or perhaps it could return 'now' for future dates? I think a date in the future is at least theoretically possible, it's hard to keep server time in sync, I've definitely seen thing like "Last updated in 2 minutes" on other platforms. So maybe we want to return the most useful response for the most likely scenario, where the date is a few seconds in the future?

mxdvl:
Or return “recently” and warn inside the console? It’s a soft fallback that isn’t wrong for the user.

oliverlloyd:
> Or return “recently” and warn inside the console? It’s a soft fallback that isn’t wrong for the user, while remaining distinctly different to other values.

I like how this provides a graceful failover in all cases 👍 . I guess the question is, do we want to trigger errors on future dates so that we discover them or handle them silently?

src/timeAgo.test.ts
mxdvl: > Or perhaps it could return 'now' for future dates? I think a date in the future is at least theoretically possible, it's hard to keep server time in sync, I've definitely seen thing like "Last updated in 2 minutes" on other platforms. So maybe we want to return the most useful response for the most likely scenario, where the date is a few seconds in the future? -@oliverlloyd 
>
> Or return “recently” and warn inside the console? It’s a soft fallback that isn’t wrong for the user, while remaining distinctly different to other values. -@mxdvl
>
> I like how this provides a graceful failover in all cases 👍 . I guess the question is, do we want to trigger errors on future dates so that we discover them or handle them silently?- @oliverlloyd 

I’m thinking we should handle silently for now? I don’t think an error in a timezones should stop JS from executing.

sndrs: it may be beyond the scope of this PR, but perhaps this module should be called `timeFromNow` or `toRelativeTime` and handle past and future dates?

sndrs (approved):
great!

src/timeAgo.test.ts
mxdvl: I think the final solution to this can be addressed in a subsequent PR :D

PR #214, opened 2021-06-18
feat: throw error if cookie name or value is invalid

arelra:
## What does this change?

Throw an error if the cookie name or value is invalid as suggested by @sndrs 

## Why?

Make errors explicit.

### DIFF
--- a/src/cookies.ts
+++ b/src/cookies.ts
@@ -1,5 +1,6 @@
 const memoizedCookies: Map<string, string> = new Map<string, string>();
 const COOKIE_REGEX = /[()<>@,;"\\/[\]?={} \t]/g;
+const ERR_INVALID_COOKIE = `Cookie must not contain invalid characters (space, tab and the following characters: '()<>@,;"/[]?={}')`;
 
 export const getCookieValues = (name: string): string[] => {
 	const nameEq = `${name}=`;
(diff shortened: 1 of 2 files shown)


src/cookies.ts
```diff
@@ -61,8 +65,8 @@ export const setCookie = ({
 }): void => {
 	const expires = new Date();
 
-	if (!isValidCookieValue(name) || !isValidCookieValue(value)) {
-		return;
+	if (!isValidCookie(name, value)) {
+		throw new Error(`${ERR_INVALID_COOKIE} .${name}=${value}`);
```
sndrs, line 69: what is the `.` doing here?

arelra: Not much I was following the format from: [cookie.ts](https://github.com/guardian/dotcom-rendering/blob/d00f45c9b6be29519b056885b0be4d818fc71b67/src/web/browser/cookie.ts#L80)

I can remove it.

src/cookies.ts
```diff
@@ -61,8 +65,8 @@ export const setCookie = ({
 }): void => {
 	const expires = new Date();
 
-	if (!isValidCookieValue(name) || !isValidCookieValue(value)) {
-		return;
+	if (!isValidCookie(name, value)) {
```
sndrs, line 68: do you think it would be worth abstracting these two check into a `validateCookie()` function that throws, rather than having two instances of the same behaviour?

i'm not sure. just a suggestion – whichever you're happy with is good with me?

sndrs (approved):
one comment which i'm totally happy for you to ignore if you want to :)

love it!

src/cookies.ts
sndrs: or maybe `throwIfInvalid()` might be more explicit

arelra: Good point! Personally I like that it's explicit that the function will not continue if the cookie is invalid. But I can see the argument to consolidate.

However overriding that is your suggestion in the previous PR, to merge `setCookie` and `setSessionCookie` so there would be only one `throw` in future anyway. I will do another PR for that.

PR #220, opened 2021-06-23
feat: migrate format from `@guardian/types`

jamie-lynch:
## What does this change?

* Migrates format from `@guardian/types` to `@guardian/libs`
* Adds `Design.Obituary` (https://github.com/guardian/types/pull/151/files)

## Why?


jamie-lynch:
cc @JamieB-gu

src/format.test.ts
```diff
@@ -0,0 +1,17 @@
...
+	expect(Display.Standard).toBeDefined();
+});
+
+it('Pillar enum contains News', () => {
+	expect(Pillar.News).toBe(0);
+});
+
+it('Special enum contains SpecialReport', () => {
+	expect(Special.SpecialReport).toBe(5);
+});
```
jamie-lynch, line 17: This keeps coveralls happy but I'm not sure that it's much use beyond that?

JamieB-gu: Ah, interesting that coveralls is really keen for you to run tests on what are essentially just type definitions - there's not really any logic to test there.

That said, testing the `Theme` enum might be useful, it's built as a compound of two other `enum`s so I guess there's a chance for something to get broken there.

Otherwise, yeah I agree, this seems reasonable just to keep coveralls happy, but it's probably just running over ground that TSC covers anyway.

src/format.ts
```diff
@@ -0,0 +1,54 @@
...
+	LiveBlog,
+	DeadBlog,
+	Recipe,
+	MatchReport,
+	Interview,
+	Editorial,
+	Quiz,
+	Interactive,
+	PhotoEssay,
+	PrintShop,
```
JamieB-gu, line 35: You may want to include `Obituary` in this list (see https://github.com/guardian/types/pull/151). Otherwise the subsequent PR to add it will trigger a major version bump 😬.

src/format.README.md
```diff
@@ -0,0 +1,10 @@
+# `format`
+
+Types and enums related to editorial formats.
+
+### Usage
```
JamieB-gu, line 5:
```suggestion
## Usage
```

JamieB-gu (commented):
Nice! Thanks for doing this 🚀

One comment about obituaries - otherwise looks good!

JamieB-gu (dismissed):

src/format.test.ts
buck06191: They're actually good tests to have. It's not unfeasible that someone might modify the type and accidentally modify an existing Type due to a typo, or  change the ordering and have that somehow break something

sndrs (dismissed):

src/format.README.md
```diff
@@ -0,0 +1,10 @@
+# `format`
+
+Types and enums related to editorial formats.
+
+## Usage
+
+```js
+import type { Theme, Format } from '@guardian/libs';
+import { Pillar, Special, Design, Display } from '@guardian/libs';
```
sndrs, line 9: i think we need to think about these names here. i know they have big significance in editorial, but examples like this:

```js
import { Special } from '@guardian/libs';
import { Design } from '@guardian/libs';
```
need to convey more to non-editorial devs i think. esp when it pops up in a list of possible imports in autocomplete (and in a generic org-wide library).

@JamieB-gu @oliverlloyd does something like this make sense?

```js
import type { EditorialTheme, EditorialFormat } from '@guardian/libs';
import { EditorialPillar, EditorialSpecial, EditorialDesign, EditorialDisplay } from '@guardian/libs';
```

JamieB-gu: I see your point. Given that `libs` is dept-wide and not Editorial-specific the names do lose some context. Perhaps this is the best solution 🤔.

Although it's tricky because by making it more specific for _non-consumers_ of these exports you're making it verbose/overly specific for the applications that _do_ consume them. I don't have an idea for how to solve that though 😔.

Given that these may not live in this library for long do we want to do the large find-and-replace PRs on DCR, AR and IR to accommodate this change?

Apologies @sndrs none of what I've said is particularly conclusive, I've just added some more questions 🙈

cc @gtrufitt

oliverlloyd: Yeah, `import { Special } from '@guardian/libs';` lacks a lot of context outside of our bubble 🤔 

`EditorialSpecial` is explicit but also a bit clunky. You'd probably end up  `import { EditorialSpecial as Special } from '@guardian/libs';` which I guess is fine but also a bit meh.

I'm guessing not but Is there an option to use a folder, like

```ts
import { Special } from '@guardian/libs/editorial';
```
?

Another option is to do what DCR does (in about 80% of places) and use the `Type` suffix, like

```ts
import { SpecialType, PillarType } from '@guardian/libs';
```

sndrs: the folder thing is an option but i'd rather avoid restructuring libs in order to accommodate importing ambiguous exports from `types`. e.g. we could potentially end up exporting two things with the same from different places which is even more ambiguity. 

how do you feel about these instead?

```ts
import type { ArticleTheme, ArticleFormat } from '@guardian/libs';
import { ArticlePillar, ArticleSpecial, ArticleDesign, ArticleDisplay } from '@guardian/libs';
```
it has the advantage of not implying that only certain teams should them and `ArticleDesign.MatchReport` reads clearly to me?

another option: if this is still unacceptably verbose, perhaps that's a sign these things shouldn't live in here?

JamieB-gu: I think I'd actually prefer `EditorialDesign` to `ArticleDesign`, because `Article` _is_ a `Design`, so you'll end up with `ArticleDesign.Article` 🙈.

> another option: if this is still unacceptably verbose, perhaps that's a sign these things shouldn't live in here?

That's reasonable. It's possible that they don't. They're specifically Editorial concepts and this is meant to be a dept-wide repo. They're only being put here for temporary convenience until we can get them into CAPI. Maybe instead we should just not deprecate `@guardian/types` until that migration to CAPI has finished?

JamieB-gu: It also occurs to me that if we _do_ merge the DCR and AR repos then having `Format` in a library becomes unnecessary. We can just put it in a shared folder in that repo.

oliverlloyd: I mean, really, `Article` should be `News` or `Standard` and then `ArticleDesign` would work really nicely 😄

sndrs: ah right that's interesting. so it sounds as though something like `EditorialContentDesign.Article` is probably an _accurate_, unambiguous name (or something like it), but the amount of context you need to pack in to that name perhaps means it's not generic enough to live in here (whereas living in CAPI provides that context implicitly)

JamieB-gu: > I mean, really, `Article` should be `News` or `Standard`

It probably should be `Standard`, you're right Oliver 🙄.

I don't think it should be `News` because a) it's not always used for news pieces, and b) `News` is a pillar 🙂.

jamie-lynch: Just to clarify, is everyone happy for these values to change to

```ts
import type { ArticleTheme, ArticleFormat } from '@guardian/libs';
import { ArticlePillar, ArticleSpecial, ArticleDesign, ArticleDisplay } from '@guardian/libs';
```

and `Design.Article` to become `ArticleDesign.Standard` or do we close this PR and figure out a better place for `Format` to live?

oliverlloyd: I'm happy with `ArticleTheme` etc. 👍 

Maybe the move from `Article` > `Standard` could be a separate PR?

oliverlloyd: But also 👍 for `Standard`

JamieB-gu (approved):

PR #221, opened 2021-06-23
feat: migrate ophan from types

jamie-lynch:
## What does this change?

* Migrates ophan from `@guardian/types` to `@guardian/libs`

## Why?

### DIFF
--- a/README.md
+++ b/README.md
@@ -71,6 +71,10 @@ Inject an external JavaScript file.
 
 Selectively log team-specific messages to the console.
 
+### [`ophan`](./src/ophan.README.md)
+
+Types related to Ophan.
+
 ### [`storage`](./src/storage.README.md)
 
 Robust API over `localStorage` and `sessionStorage`.
--- a/README.md
+++ b/README.md
@@ -21,6 +21,7 @@
   - [`isUndefined`](#isundefined)
   - [`loadScript`](#loadscript)
   - [`log`/`debug`](#logdebug)
+  - [`ophan`](#ophan)
   - [`storage`](#storage)
   - [`timeAgo`](#timeago)
 - [Installation](#installation)
--- a/src/index.ts
+++ b/src/index.ts
@@ -14,6 +14,16 @@ export { isString } from './isString';
 export { isUndefined } from './isUndefined';
 export { loadScript } from './loadScript';
 export { debug, log } from './logger';
+export type {
+	OphanABEvent,
+	OphanABPayload,
+	OphanAction,
+	OphanComponent,
+	OphanComponentEvent,
+	OphanComponentType,
+	OphanProduct,
+	TestMeta,
+} from './ophan';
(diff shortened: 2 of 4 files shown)


jamie-lynch:
This only provides types so should this be in the `src/types` directory instead?

README.md
```diff
@@ -71,6 +72,10 @@ Inject an external JavaScript file.
 
 Selectively log team-specific messages to the console.
 
+### [`ophan`](./src/ophan.README.md)
+
+Types related to Ophan.
```
jamie-lynch, line 77: Suggestions for a more useful summary welcome!

src/index.ts
```diff
@@ -16,4 +16,14 @@ export { loadScript } from './loadScript';
...
 export { timeAgo } from './timeAgo';
+export type {
+	OphanABEvent,
+	OphanABPayload,
+	OphanAction,
+	OphanComponent,
+	OphanComponentEvent,
+	OphanComponentType,
+	OphanProduct,
+	TestMeta,
```
sndrs, line 27: @tjmw can you explain what `TestMeta` is?

the other ophan types are really nicely named, but this is a bit confusing. is it ophan specific or more generic, or specific to AB test?

tjmw: Hi @sndrs. Yeah that name is a bit overly generic isn't it. I think it's additional data that gets sent to Ophan to provide context related to an A/B test. I spent a moment looking back in the Git history but it looks like it's been that way for a while. I extracted this stuff from DCR, you can see the PRs where we removed from DCR and started using the `@guardian/types` types [here](https://github.com/guardian/types/pull/24) and [here](https://github.com/guardian/dotcom-rendering/pull/1953). I wonder what a better name would be to convey the purpose of this 🤔 .

sndrs: ah okedoke, sorry i just pulled your name from the blame :)

maybe `OphanABTestMeta` then? def looks like it's ophan-specific now i look again:

```js
componentType: OphanComponentType;
products?: OphanProduct[];
```

tjmw: Yeah something like that sounds good 👍  It's definitely Ophan specific as you say. I suppose the only question is do you make the (breaking) name change at the same time as moving these types from `types` -> `libs`. Or move them over and keep them identical, following up with the name change separately.

jamie-lynch: We've made (or are making) some other changes in the process of migrating other parts of `types` over to `libs`. I _think_ it's better that way so that you can make a couple of find-and-replace changes at the point you bump the version  of `libs` and remove `types`, rather than having to bump `libs` multiple times and make the changes one at a time. Having said that, I'd rather go with what everyone (or at least the majority of people) that consume `libs` think is the best approach.

I'll change this name for now but can always revert if we decide to move as-is and update later.

sndrs (approved):
great!

PR #222, opened 2021-06-23
feat: migrate role from types

jamie-lynch:
## What does this change?

* Migrates role from `@guardian/types` to `@guardian/libs`

## Why?


README.md
```diff
@@ -71,6 +72,10 @@ Inject an external JavaScript file.
 
 Selectively log team-specific messages to the console.
 
+### [`role`](./src/role.README.md)
+
+<!-- Insert summary here -->
```
jamie-lynch, line 77: I wasn't able to guess at a useful summary here so suggestions welcome.

src/role.test.ts
```diff
@@ -0,0 +1,5 @@
+import { Role } from './role';
+
+it('Role enum contains Standard', () => {
+	expect(Role.Standard).toBeDefined();
+});
```
jamie-lynch, line 5: Similar to https://github.com/guardian/libs/pull/220#discussion_r657176486, I'm not sure if this is useful beyond keeping coveralls happy.

README.md
sndrs: We should def rename this :)

sndrs: I have no idea what to though

src/role.test.ts
mxdvl: At the very least it means that if you’re changing a type, you need to change it in two places, which seems like it’s a good reminded that it is a breaking change?

README.md
jamie-lynch: @JamieB-gu Looks like it might have been you that added this initially (https://github.com/guardian/types/pull/33). We're not entirely sure what this is / what it's used for so have you got any suggestions for the summary and a more specific name here?

JamieB-gu: I can give it a go.

It's a semantic/layout-oriented property that's applied to elements in an article. It describes the "role" of that element in the piece and how it should be laid out within the page. It's probably easiest to understand when thinking about images. Is the image filling a "supporting" role? Is it just a thumbnail (for example the author avatars [here](https://www.theguardian.com/commentisfree/2020/mar/11/2020-budget-guardian-panel-verdict-rishi-sunak))? Is it "showcasing" something (as in [this piece](https://www.theguardian.com/film/2018/dec/27/best-films-2018-you-didnt-see-unsane-cielo)). Whilst most commonly used for images, it can also apply to atoms and embeds.

That's the limit of my understanding, but @paperboyo can probably give a more accurate answer with some historical context.

I don't think it has another name; "role" is how it's referred to [throughout the CAPI models](https://github.com/guardian/content-api-models/blob/master/models/src/main/thrift/content/v1.thrift) and most likely in Composer too.

paperboyo: All what Jamie said is true! Only Composer calls it `weighting` in the UI, because nothing is ever perfect.

jamie-lynch: Great, thanks for explaining that. In that case, could we call it something like `ArticleElementRole` to make it a bit more explicit? In the meantime, I'll update the README to add this in.

JamieB-gu: The argument @sndrs made about `Format` not belonging in `libs` is perhaps also relevant to `Role`. If it's specifically an Editorial concept that's going to lose context when placed in a "generic" repo like `libs` then perhaps this is ultimately not the place for it?

jamie-lynch: Yeah, that's a fair point. I've add a comment to the thread in #220 so once we reach a final decision for `Format`, should we just do the same thing with `Role`?

JamieB-gu (approved):

Issue #227, opened 2021-06-28
docs are becoming unwieldy


sndrs:
1. move module-level READMEs to JSDOC comments in modules
2. auto-build main readme from JSDOC

mxdvl:
On the subject of auto-building readme files: is it worth having the semi-automated table of contents in the README file? Github offers one out of the box already?

![duplicate table of contents](https://user-images.githubusercontent.com/76776/124141729-a3cc0980-da57-11eb-8660-36f78fb9cf25.png)

PR #237, opened 2021-07-06
feat: Added `joinUrl` module

oliverlloyd:
## What does this change?
Adds the `joinUrl` function as a new module

## Why?
This is used in DCR to prevent errors with missing or double slashes when concatenating strings to form urls and would be useful for other projects.


oliverlloyd:
[`joinUrl` in @guardian/libs](https://trello.com/c/8s9ZgrLP/3186-joinurl-in-guardian-libs)

src/joinUrl.test.ts
```diff
@@ -0,0 +1,45 @@
+import { joinUrl } from './joinUrl';
+
+const expectedOutput = 'http://example.com/abc/xyz';
+
+describe('joinUrl', () => {
+	it('prevents double slashes correctly', () => {
+		expect(joinUrl('http://example.com/', '/abc/', '/xyz/')).toBe(
+			expectedOutput,
```
tjmw, line 8: Totally personal preference but I'd move the `expectedOutput` const into each test case rather than sharing (or even inline the string). I love not having to bounce around the test file to fully understand a test, and I'd argue it makes it easier to change a single test down the line without having to worry about impact on other tests.

tjmw (approved):
This looks great! I left one (very opinionated) suggestion :-)

mxdvl (commented):
This looks like a useful method!

I’d really like to know what the result of `joinUrl('//example.com/','index.html')` returns, as I fear it might not support an [implicit protocol URL][1], starting with double-slashes.

Worth noting that we should probably generate some kind of warning for non-SSL pages? Paul Irish mentioned this was an anti-pattern [back in 2014].

[1]: https://developer.mozilla.org/en-US/docs/Learn/Common_questions/What_is_a_URL#examples_of_absolute_urls
[back in 2014]: https://www.paulirish.com/2010/the-protocol-relative-url/

README.md
```diff
@@ -24,6 +24,7 @@
   - [`log`/`debug`](#logdebug)
   - [`storage`](#storage)
   - [`timeAgo`](#timeago)
+  - [`joinUrl`](#joinurl)
```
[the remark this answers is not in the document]
sndrs: might be some others instances? (sorry, we need to sort these docs out, getting a bit unwieldy #227)

sndrs (changes requested):

src/joinUrl.test.ts
oliverlloyd: I wholeheartedly agree, tests should be as simple as possible. Thanks for pointing this out, I'll update

src/joinUrl.ts
```diff
@@ -0,0 +1,18 @@
...
+			// Trim left
+			if (part.substr(0, 1) === '/') return part.slice(1);
+			return part;
+		})
+		.map((part) => {
+			// Trim right
+			if (part.substr(part.length - 1, 1) === '/')
+				return part.slice(0, -1);
+			return part;
+		});
```
sndrs, line 15: could you replace both of these with `.map(part => part.replace(/^\/|\/$/g, ''))`?

not sure what the perf implications are though, and only a style thing so not a blocker

oliverlloyd: I could... but I fear that would leave the code harder to read - or at least to me it reads as a series of slashes and symbols.

I tend to reach for regexes when I _need_ them or if there's a strong performance gain from using them. Outside of this, I find javascript's own native methods a nicer set of tools to use.

That said, I'm not against them, maybe I just need to learn them better? And I'm not strongly opposed to changing this code here. It's also interesting that you raise it as a style consideration which means that not everyone has the same subjective ideas as me when it comes to how code should look (shocker!)

oliverlloyd:
> I’d really like to know what the result of `joinUrl('//example.com/','index.html')` returns, as I fear it might not support an [implicit protocol URL](https://developer.mozilla.org/en-US/docs/Learn/Common_questions/What_is_a_URL#examples_of_absolute_urls), starting with double-slashes.

```
	it('does not work with implicit protocol urls', () => {
		expect(joinUrl('//example.com/', '/index.html')).toBe(
			'/example.com/index.html',
		);
	});
```

I can add this test as an indication of the fact it won't support this 🤷‍♂. I can't think of a way to know when you'd want to preserve leading double slashes

oliverlloyd:
> Worth noting that we should probably generate some kind of warning for non-SSL pages?

Interesting point. I wonder if this kind of check would work better as a Github action higher up? Ot maybe a lint rule?

src/joinUrl.ts
sndrs: yeah if you prefer it this way all good 👍

sndrs: > maybe I just need to learn them better?

things of beauty :)

sndrs: also regex way was 34% slower on jsbench.me so this has just been noise. apologies!

src/joinUrl.ts
```diff
@@ -0,0 +1,18 @@
+export const joinUrl = (...args: string[]): string => {
+	// Remove any leading or trailing slashes from all parts and then join cleanly on
+	// a single slash, preventing malformed urls
+	const trimmed = args
+		.map((part) => {
+			// Trim left
+			if (part.substr(0, 1) === '/') return part.slice(1);
```
sndrs, line 7:
```suggestion
			if (part.startsWith('/')) return part.slice(1);
```

src/joinUrl.ts
```diff
@@ -0,0 +1,18 @@
...
+			return part;
+		})
+		.map((part) => {
+			// Trim right
+			if (part.substr(part.length - 1, 1) === '/')
+				return part.slice(0, -1);
```
sndrs, line 13:
```suggestion
			if (part.endsWith('/'))	return part.slice(0, -1);
```

sndrs (commented):
sorry, another suggestion. feel free to dismiss if you dont like them

sndrs:
> I can add this test as an indication of the fact it won't support this 🤷‍♂. I can't think of a way to know when you'd want to preserve leading double slashes

should it throw if it encounters one in this case? that way you can throw a meaningful error rather than return an invalid string?

oliverlloyd:
> ```ts
> // detect two or more `/` chars after a word boundary
> const multipleSlashesInRoute = /\b\/{2,}/g; 
> 
> export const joinUrl = (...args: string[]): string =>
> 	args.join('/').replace(multipleSlashesInRoute, '/');
> ```

This works beautifully! It achieves the same results but also allows for leading double slashes and allows for a trailing slash. And it does all this in a really simple, readable fashion (even as a regex 😉 )

Really great suggestion

src/joinUrl.ts
oliverlloyd: No longer needed (but yes, this is much better)

src/joinUrl.ts
oliverlloyd: No longer needed

sndrs (approved):

mxdvl:
From [Regex101](https://regex101.com/r/kl6yUC/1) on `/\b\/{2,}/g` :

> `\b` assert position at a word boundary: `(^\w|\w$|\W\w|\w\W)`
> `\/` matches the character `/` literally _(case sensitive)_
> `{2,}` matches the previous token between `2` and `unlimited` times, as many times as possible, giving back as needed _(greedy)_
> `g` modifier: `global`. All matches _(don't return after first match)_

akash1810:
Could we use the native [`URL`](https://developer.mozilla.org/en-US/docs/Web/API/URL/URL#examples) object instead (with the added benefit of a strongly typed objects)?

PR #274, opened 2021-09-07
feat: Add `sendJson`

mxdvl:
## What does this change?

Adds a way to safely send a JSON payload, using `navigator.sendBeacon` with a `fetch` fallback.

## Why?

`sendBeacon` isn’t fully supported.


mxdvl:
Actually, I think it is better to [polyfill `sendBeacon`](https://github.com/guardian/frontend/pull/24164)

PR #280, opened 2021-09-13
feat: Add `coreWebVitals` method

mxdvl:
## What does this change?

Brings Core web vitals out of frontend and into `libs`, so it can be shared and kept in sync between projects.

## Why?

Some improvements were needed in frontend, and we needed a consitent API between various platforms.

That’s what libs are for.


nitro-marky (approved):
Nice

src/coreWebVitals.ts
```diff
@@ -0,0 +1,121 @@
...
+	page_view_id: string | null;
+	browser_id: string | null;
+	fid: null | number;
+	cls: null | number;
+	lcp: null | number;
+	fcp: null | number;
+	ttfb: null | number;
+};
+
+const coreWebVitalsPayload: CoreWebVitalsPayload = {
```
zekehuntergreen, line 14: It might be worth giving this a different name for clarity but I don't have a great suggestion.
Maybe the type is `CoreWebVitalsPayload` and the const is `pageCoreWebVitals`?

src/coreWebVitals.ts
```diff
@@ -0,0 +1,121 @@
...
+			coreWebVitalsPayload.fid = roundWithDecimals(metric.value);
+			break;
+		case 'TTFB':
+			// Browser support: Chromium, Firefox, Safari, Internet Explorer
+			coreWebVitalsPayload.ttfb = roundWithDecimals(metric.value);
+			break;
+	}
+};
+
+export const coreWebVitals = (
```
zekehuntergreen, line 81: This is an improvement on the current frontend naming but maybe it could be made even more clear.
`registerCoreWebVitalHandlers` ? `registerCoreWebVitalCallbacks`?

src/coreWebVitals.ts
```diff
@@ -0,0 +1,121 @@
...
+		return false;
+
+	const endpoint =
+		window.location.hostname === 'm.code.dev-theguardian.com' ||
+		window.location.hostname === 'localhost' ||
+		window.location.hostname === 'preview.gutools.co.uk'
+			? 'https://performance-events.code.dev-guardianapis.com/core-web-vitals'
+			: 'https://performance-events.guardianapis.com/core-web-vitals';
+
+	return navigator.sendBeacon(endpoint, JSON.stringify(coreWebVitalsPayload));
```
zekehuntergreen, line 53: Have we decided not to make the change from `sendBeacon` to `fetch`? I thought I understood that `fetch` had wider support.

src/coreWebVitals.ts
```diff
@@ -0,0 +1,121 @@
...
+	{
+		browserId,
+		pageViewId,
+		forceSendMetrics = false,
+	}: {
+		browserId: string;
+		pageViewId: string;
+		forceSendMetrics: boolean;
+	},
+	metricsSentCallback: (queued?: boolean) => void = () => void 0,
```
zekehuntergreen, line 91: Do we have a use case for this callback already? If I understand right it's not being used at the moment, right?

src/coreWebVitals.ts
```diff
@@ -0,0 +1,121 @@
...
+			const queued = sendData();
+			metricsSentCallback(queued);
+		}
+	});
+	// Safari does not reliably fire the `visibilitychange` on page unload.
+	addEventListener('pagehide', () => {
+		const queued = sendData();
+		metricsSentCallback(queued);
+	});
+};
```
zekehuntergreen, line 116: This approach seems like a big improvement - sending a payload which is as populated as possible once per page rather than sending cumulatively populated payload many times.
Looks like it's also the recommended way to send analytics data https://developer.mozilla.org/en-US/docs/Web/API/Document/visibilitychange_event#usage_notes

src/coreWebVitals.ts
```diff
@@ -0,0 +1,121 @@
...
+	const power = Math.pow(10, precision);
+	return Math.round(value * power) / power;
+};
+
+const sendData = (): boolean => {
+	if (
+		!(userInSample || shouldSendMetrics) ||
+		coreWebVitalsPayload.fcp === null
+	)
+		return false;
```
zekehuntergreen, line 44: This might be out of scope of the pr but could we do this even earlier, at the top of `coreWebVitals`?
No point in registering the web-vitals callbacks if we aren't going to end up sending the results to the lake, right?

zekehuntergreen (commented):
+1 for unifying the very similar DCR and frontend code!

src/coreWebVitals.ts
mxdvl: So in the future, this will help us implement `log` calls.

src/coreWebVitals.ts
mxdvl: Yes, I was going to add a note about this.

It’s also recommended by `web-vitals` https://github.com/GoogleChrome/web-vitals#batch-multiple-reports-together

src/coreWebVitals.ts
mxdvl: `initCoreWebVitals` ?

src/coreWebVitals.ts
mxdvl: We’ve added a polyfill for `navigator.sendBeacon` on [frontend] and [DCR].

[frontend]: https://github.com/guardian/frontend/pull/24164
[DCR]: https://github.com/guardian/dotcom-rendering/pull/3415

src/coreWebVitals.ts
mxdvl: I think the `camelCase: PascalCase` notation for variables and their types is quite idiomatic in JS and Scala codebases.

src/coreWebVitals.ts
mxdvl: You have to register the callbacks if you want to be able to asynchronously force metrics to be sent. Otherwise you’ll only get partial data, and TTFB and FCP are likely to be missing.

src/coreWebVitals.README.md
```diff
@@ -0,0 +1,72 @@
...
+import { initCoreWebVitals } from '@guardian/libs';
+
+const optionalCallback = () => {
+    // An optional callback which triggers when the user leaves the page
+    log('dotcom', 'CWV payload queued for async request')
+}
+
+initCoreWebVitals({
+    browserId,
+    pageViewId,
```
oliverlloyd, line 23: I can imagine future developers that are not part of the Dotcom team might wonder how these values can be generated/read. Is it worth making a note about this here?

src/coreWebVitals.README.md
```diff
@@ -0,0 +1,72 @@
...
+    browserId: 'abc',
+    pageViewId: '123',
+    isDev: false,
+    forceSendMetrics: isInSeverSideTest(),
+}
+
+initCoreWebVitals(init)
+```
+
+### `forceSendMetrics`
```
oliverlloyd, line 44: I find this confusing. Saying 'force' implies that it's an optional way to send the data in special circumstances and that simply calling  `initCoreWebVitals` is enough - is that true? Or do I _need_ to call this to have the metrics sent?

src/coreWebVitals.test.ts
```diff
@@ -0,0 +1,219 @@
...
+		initCoreWebVitals({ browserId, pageViewId });
+
+		expect(sendData()).toBe(true);
+	});
+
+	it('should not send data if not in sample', () => {
+		initCoreWebVitals({ browserId, pageViewId });
+
+		expect(sendData()).toBe(false);
+	});
```
oliverlloyd, line 170: What's the difference between these two tests? I can see one has a mock for a call which means the metrics are not sent, but other one expects the results to never be sent but for no obvious reason

Is it possible to make it more explicit why one test expects one things and the other as part of the test code itself?

src/coreWebVitals.test.ts
```diff
@@ -0,0 +1,219 @@
...
+
+	it('should send data if not in sample but forced via init', () => {
```
oliverlloyd, line 172: Did you mean to include 'not in sample' in the description here?

src/coreWebVitals.test.ts
```diff
@@ -0,0 +1,219 @@
...
+	it('should send data if forced asynchronously', () => {
+		initCoreWebVitals({ browserId, pageViewId });
+
+		expect(sendData()).toBe(false);
+
+		forceSendMetrics();
+		expect(sendData()).toBe(true);
+	});
+
+	it('should use CODE URL by default', () => {
```
oliverlloyd, line 198: Should it? Is there a way to make specifying the endpoint required so there's no default? I feel like if you are falling back on a default something has gone wrong

src/coreWebVitals.test.ts
```diff
@@ -0,0 +1,219 @@
...
+		forceSendMetrics();
+
+		sendData(false);
+		expect(mockBeacon).toHaveBeenCalledWith(
+			'https://performance-events.guardianapis.com/core-web-vitals',
+			expect.any(String),
+		);
+	});
+
+	it('should use PROD URL if isDev is false', () => {
```
oliverlloyd, line 209: How can I tell `isDev` is false here?

src/coreWebVitals.ts
```diff
@@ -0,0 +1,127 @@
...
+export const initCoreWebVitals = (
+	{
+		browserId,
+		pageViewId,
+		forceSendMetrics = false,
+		isDev,
+	}: {
+		browserId: string;
+		pageViewId: string;
+		isDev?: boolean;
```
oliverlloyd, line 85: I alluded to this earlier but perhaps this should be required?

src/coreWebVitals.ts
```diff
@@ -0,0 +1,127 @@
...
+ * @param init - the initialisation options
+ * @param init.isDev - Whether to log to CODE or PROD endpoints
+ * @param metricsSentCallback - Optional callback, triggered after metrics are queued for sending
+ */
+export const initCoreWebVitals = (
+	{
+		browserId,
+		pageViewId,
+		forceSendMetrics = false,
+		isDev,
```
oliverlloyd, line 81: The contra to my other comment, if we _don't_ make this required then can we default it here rather than later on?

src/coreWebVitals.ts
```diff
@@ -0,0 +1,127 @@
...
+	coreWebVitalsPayload.browser_id = browserId;
+	coreWebVitalsPayload.page_view_id = pageViewId;
+
+	// By default, sample 1% of page views
+	const pageViewInSample = Math.random() < 1 / 100;
+	// Unless we are forcing sending metrics for this page view
+	// via initialisation or calling forceSendMetrics()
+	if (forceSendMetrics || pageViewInSample) shouldSendMetrics = true;
+	// Or using a specific hash
+	if (window.location.hash === '#forceSendMetrics') shouldSendMetrics = true;
```
oliverlloyd, line 99: Well done, this code looks really simple but I can see how it could have been confusing but you've avoided that and done it really well 👍

oliverlloyd (dismissed):
Impressive stuff. I'd be keen to start using this on some pages.

I made some comments as someone coming at this with very little context or background knowledge, mostly ideas to hopefully make things more readable for others.

src/coreWebVitals.README.md
nitro-marky: If you have access to BigQuery they can be seen as raw data there but we have a dashboard here which can be split into DCR and frontend if you go to `Filtering Section` and change `Filter Component`  https://tableau.gnmremote.com/#/views/DotcomWebVitals/Singlefilterview/guy_parr@gnm.int/397372d8-b69d-44e8-acaf-45c60de7c4ed?:display_count=n&:showVizHome=n&:origin=viz_share_link&:iid=1

src/coreWebVitals.README.md
nitro-marky: +1. If they are part of the 1% group nothing is forced as they have been selected, so `init` will take place.

src/coreWebVitals.README.md
oliverlloyd: Sorry, by 'read' I meant should I read them from a cookie, from local storage, or do I need to generate new values myself? There's some assumed knowledge built in here that not everyone looking to add the ability to track their core web vitals will have. They could ask around, sure, but if we make it easier for them it might improve usage and make the world a slightly faster place.

mxdvl: That’s a good point, @oliverlloyd. Actually, I think that these params should be optional, because you can still get aggregate data, even if you can’t join it to page views. I’ll also add an example of where they usually can be found and some documentation.

src/coreWebVitals.README.md
mxdvl: Another good point. This is meant to bypass the sampling value. With your comments, I’m thinking of changing the init params to make this more explicit, and renaming this to `bypassSampling`.

mxdvl:
Gathering some feedback, I’m thinking of changing the initialisation options.

```ts
type InitCoreWebVitalsOptions = {
	browserId?: string | null; 
	pageViewId?: string | null; // if unset, data can’t be joined
	isDev: boolean; // now required
	sampling?: number; // defaults to 1%
	bypassSampling?: boolean; // if true, same as setting sampling to 100%
};
```

src/coreWebVitals.ts
mxdvl: Thanks for the kind words. I think your input has made it even nicer. 🌷

src/coreWebVitals.test.ts
mxdvl: Updated now. Somehow I can’t resolve this converstaion.

oliverlloyd (dismissed):
👍 ❤️

src/coreWebVitals.ts
```diff
@@ -0,0 +1,160 @@
...
+ * @param init.sampling - sampling rate for sending data. Defaults to `0.01`.
+ * @param init.bypassSampling - send data regardless of sampling rate. Defaults to `false`.
+ *
+ * @param metricsSentCallback - Optional callback, triggered after metrics are queued for sending
+ */
+export const initCoreWebVitals = (
+	{
+		browserId = null,
+		pageViewId = null,
+		bypassSampling = false,
```
sndrs, line 102: this can be handled by setting `{sampling: 1}` - is it needed? 

(in theory you could set `{sampling:0.1, bypassSampling:true}` which feels weird)

src/coreWebVitals.ts
```diff
@@ -0,0 +1,160 @@
...
+	if (sampling === 0) console.warn('Core Web Vitals are sampled at 0%');
+	if (sampling === 1) console.warn('Core Web Vitals are sampled at 100%');
+
+	// By default, sample a percentage of page views
+	const pageViewInSample = Math.random() < sampling;
+	// Unless we are forcing sending metrics for this page view
+	// via initialisation or calling bypassSampling()
+	if (bypassSampling || pageViewInSample) shouldSendMetrics = true;
+	// Or using a specific hash
+	if (window.location.hash === '#bypassSampling') shouldSendMetrics = true;
```
sndrs, line 132: `#bypassSampling` is a quite generic, how about this?

```suggestion
	if (window.location.hash === '#bypassCVWSampling') shouldSendMetrics = true;
```

src/coreWebVitals.ts
```diff
@@ -0,0 +1,160 @@
...
+	addEventListener('visibilitychange', () => {
+		if (document.visibilityState === 'hidden') {
+			const queued = sendData(isDev);
+			if (metricsSentCallback) metricsSentCallback(queued);
+		}
+	});
+	// Safari does not reliably fire the `visibilitychange` on page unload.
+	addEventListener('pagehide', () => {
+		const queued = sendData(isDev);
+		if (metricsSentCallback) metricsSentCallback(queued);
```
sndrs, line 150: if you have more than one instance of this module on a page, you would need a way of subscribing multiple times. maybe a `onCoreWebVitalsSent` export?

src/coreWebVitals.ts
nitro-marky: In case my comment above is missed we absolutely need data that ties it to a page view. Otherwise it becomes worthless in some cases.

PR #288, opened 2021-09-28
chore: Add validateSingleCommit option to pr checker

jamie-lynch:
## What does this change?

- Sets the `validateSingleCommit` option to true in the PR title checker workflow

## Why?

When using "Squash and merge" on a PR with only one commit, GitHub will suggest using that commit message instead of the PR title for the merge commit, and it's easy to commit this by mistake. Enable this option to also validate the commit message for one commit PRs.


mxdvl (approved):
That’s a good one that’s saved us many times!

PR #294, opened 2021-11-05
fix: update JSDoc to match documentation

mxdvl:
## What does this change?

Matches JSDoc to documentation

## Why?

@oliverlloyd has mentioned this is the proper way


SiAdcock (approved):
👍

PR #296, opened 2021-11-23
feat: Added `whenIdle`

oliverlloyd:
## What?
`whenIdle` is a function that takes a callback which is executed either:

when `window.requestIdleCallback` runs
when the 500ms timeout is reached
or, if `requestIdleCallback` is not supported, after 300ms

## Why?
This offers people a utility for deferring code execution.

This function is proposed in [this PR for React Islands](https://github.com/guardian/dotcom-rendering/pull/3629) but it is also something that could be useful in the work to [defer downloading YouTube javascript](https://github.com/guardian/atoms-rendering/pull/303).

### DIFF
--- a/src/whenIdle.test.ts
+++ b/src/whenIdle.test.ts
@@ -1,0 +1,23 @@
+import { whenIdle } from './whenIdle';
+
+describe('whenIdle', () => {
+	it('does not fire callback straight away', () => {
+		let fired = false;
+		const cb = () => {
+			fired = true;
+		};
+		whenIdle(cb);
+		expect(fired).toBe(false);
+	});
+
+	it('fires the callback after 600ms', async () => {
+		let fired = false;
+		const cb = () => {
+			fired = true;
+		};
+		whenIdle(cb);
+		expect(fired).toBe(false);
+		await new Promise((r) => setTimeout(r, 600));
+		expect(fired).toBe(true);
+	});
+});
--- a/src/whenIdle.ts
+++ b/src/whenIdle.ts
@@ -1,0 +1,12 @@
+/**
+ * whenIdle executes the given callback when the browser is 'idle'
+ *
+ * @param callback Fired when requestIdleCallback runs or, if requestIdleCallback is not available, after 300ms
+ */
+export const whenIdle = (callback: () => void): void => {
+	if ('requestIdleCallback' in window) {
+		window.requestIdleCallback(callback, { timeout: 500 });
+	} else {
+		setTimeout(callback, 300);
+	}
+};
--- a/src/whenIdle.ts
+++ b/src/whenIdle.ts
@@ -2,10 +2,19 @@
  * whenIdle executes the given callback when the browser is 'idle'
  *
  * @param callback Fired when requestIdleCallback runs or, if requestIdleCallback is not available, after 300ms
+ * @param options Options for requestIdleCallback
+ * @param options.timeout How long to wait for requestIdleCallback to return, defaults to 500ms
  */
-export const whenIdle = (callback: () => void): void => {
+export const whenIdle = (
+	callback: () => void,
+	options: {
+		timeout: number;
+	} = {
+		timeout: 500,
+	},
+): void => {
 	if ('requestIdleCallback' in window) {
-		window.requestIdleCallback(callback, { timeout: 500 });
+		window.requestIdleCallback(callback, options);
 	} else {
 		setTimeout(callback, 300);
 	}
--- a/README.md
+++ b/README.md
@@ -104,6 +104,10 @@ Turns absolute dates into time ago strings.
 
 Combines separate url parts into one valid url string.
 
+### [`whenIdle`](./src/whenIdle.README.md)
+
+Defers code execution until the browser is idle.
+
 ## Installation
 
 [![Generic badge](https://img.shields.io/badge/google-chat-259082.svg)](https://chat.google.com/room/AAAAWwBdSMs)
--- a/README.md
+++ b/README.md
@@ -28,6 +28,7 @@
   - [`storage`](#storage)
   - [`timeAgo`](#timeago)
   - [`joinUrl`](#joinurl)
+  - [`whenIdle`](#whenidle)
 - [Installation](#installation)
   - [Bundling](#bundling)
 - [Development](#development)
--- a/package.json
+++ b/package.json
@@ -42,6 +42,7 @@
     "@rollup/plugin-typescript": "8.2.5",
     "@semantic-release/github": "7.2.3",
     "@types/jest": "27.0.1",
+    "@types/requestidlecallback": "0.3.4",
     "@types/wcag-contrast": "3.0.0",
     "conventional-changelog-conventionalcommits": "4.6.0",
     "eslint": "7.32.0",
--- a/src/whenIdle.README.md
+++ b/src/whenIdle.README.md
@@ -1,0 +1,15 @@
+# `whenIdle()`
+
+Returns: `void`
+
+Runs the given callback when it detects that the browser is idle or if 500ms has passed
+
+## Example
+
+```js
+import { whenIdle } from '@guardian/libs';
+
+whenIdle(function () {
+    console.log('The browser is idle');
+});
+```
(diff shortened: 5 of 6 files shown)


src/whenIdle.test.ts
```diff
@@ -0,0 +1,23 @@
...
+	});
+
+	it('fires the callback after 600ms', async () => {
+		let fired = false;
+		const cb = () => {
+			fired = true;
+		};
+		whenIdle(cb);
+		expect(fired).toBe(false);
+		await new Promise((r) => setTimeout(r, 600));
```
SiAdcock, line 20: Have you tried using [Jest's timer mocks](https://jestjs.io/docs/timer-mocks) for this kind of behaviour? I expect it's more deterministic.

src/whenIdle.ts
```diff
@@ -0,0 +1,12 @@
+/**
+ * whenIdle executes the given callback when the browser is 'idle'
+ *
+ * @param callback Fired when requestIdleCallback runs or, if requestIdleCallback is not available, after 300ms
+ */
+export const whenIdle = (callback: () => void): void => {
+	if ('requestIdleCallback' in window) {
+		window.requestIdleCallback(callback, { timeout: 500 });
```
SiAdcock, line 8: There is a case to be made for allowing the timeout to be overridable as part of the API (defaulting to 500ms). What do you think?

oliverlloyd: `options`!

src/whenIdle.test.ts
oliverlloyd: Those new mocks for Jest look nice. The way Jest used globals before was painful. However, in this case I'm not sure it works. I need to advance _real_ time, not a mocked `setTimeout`, because the code is - potentially - using the idle callback. I guess you could argue we could just ignore the actual idle event and force the setTimeout but maybe that's less coverage?

sndrs:
what's the benefit of this over using `requestIdleCallback` natively?

src/whenIdle.ts
```diff
@@ -0,0 +1,21 @@
...
+	options: {
+		timeout: number;
+	} = {
+		timeout: 500,
+	},
+): void => {
+	if ('requestIdleCallback' in window) {
+		window.requestIdleCallback(callback, options);
+	} else {
+		setTimeout(callback, 300);
```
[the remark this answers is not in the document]
mxdvl: yes, you probably want to use `options.timeout`, and update the JSDoc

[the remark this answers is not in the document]
oliverlloyd: The thinking is we timeout at an absolute maximum of 500ms when `requestIdleCallback` is supported but use a sort of average (300ms) when it's not. These values will always be arbitrary but there [was some discussion](https://github.com/guardian/dotcom-rendering/pull/3629#discussion_r749286392) behind them

oliverlloyd:
> what's the benefit of this over using `requestIdleCallback` natively? the timeout?

It's mostly the api. I like how I can write code like

```ts
			switch (when) {
				case 'idle': {
					whenIdle(() => {
						doHydration(name, props, marker);
					});
					break;
				}
				case 'visible': {
					whenVisible(marker, () => {
						doHydration(name, props, marker);
					});
					break;
				}
				case 'immediate':
				default: {
					doHydration(name, props, marker);
				}
			}
```

Bit it also removes the need for people to have to know that this is an available option. By having a suite of these sorts of defer when functions we give people the tools to construct good deferred experiences without needing to do the research.

SiAdcock:
I think with `requestIdleCallback`, the `timeout` is optional but it's also best practice to provide one. So this abstraction removes the need for developers to know the best practice.

> without needing to do the research

I know that you mean "without needing to know about the timeout" in this case. But the advantage of requiring devs to do the research (i.e. by _not_ providing this API) is that they would have to look at MDN and other sources of information that discuss _when_ you should be using `requestIdleCallback` (i.e. for low priority tasks that may never actually get run, and you're okay with that fact). By providing this API, the burden would be upon us to reiterate that advice and ensure we keep it up to date.

oliverlloyd:
> But the advantage of requiring devs to do the research (i.e. by _not_ providing this API) is that they would have to look at MDN and other sources of information that discuss _when_ you should be using `requestIdleCallback`

On the other hand, the disadvantage to not providing the api is devs might not know it's an option. I didn't know about it until I saw it in the Astro api. There's also the risk that by not providing it in the api you send the message that it's been excluded, or isn't approved.

Searching around. I see we have previously abstracted this function [once before](https://github.com/guardian/grid/blob/e669dfc4b01aa4fdfa82893d535e4a86558319b3/kahuna/public/js/util/idleTimeout.js) and used in inline [once before](https://github.com/guardian/frontend/blob/5b970cd7308175cfc1bcae2d4fb8c06ee13c5fa0/static/src/javascripts/bootstraps/standard/atoms.js#L17). Arguably, they were the both the right choice in their own contexts. I'm not sure if either of them give or take weight from this PR 🤷 

I do agree that it's good to encourage devs to do their own research but in this case if the only information is about when to employ this method I think that could be achieved better in the context of a suite of options where you're explaining which one to choose. The assumption is the developer has already had the thought they want to defer execution, the decision is more about how.

Perhaps this all makes more sense if we share the same vision. In my head, I'm imagining a situation where we provide a suite of functions or patterns to defer execution of code. Currently, we're proposing doing this using [Islands](https://github.com/guardian/dotcom-rendering/pull/3629) but maybe there are other places where we can implement the same patterns but perhaps they've not been employed because it's never been considered?

Or, and this is possible, maybe I'm guessing future use cases that don't exist based on assumptions that are false and we don't need this abstraction here? I'm perfectly happy with this outcome!

mxdvl:
Just wanted to chime in with [Polyfill.io’s source](https://github.com/Financial-Times/polyfill-library/tree/master/polyfills/requestIdleCallback). It’s _chunky_.

SiAdcock:
>  In my head, I'm imagining a situation where we provide a suite of functions or patterns to defer execution of code.

I must admit this is super compelling 😄 I like the vision of the suite of functions you're imagining and I believe people would enjoy thinking in terms of `visible | idle | whenever`. I wonder if there's a way to organise this suite of functions to ensure people understand the available options... Would `deferUntilIdle` and `deferUntilVisible` send a stronger signal of relatedness?

src/whenIdle.test.ts
SiAdcock: Yes you're right, it doesn't look like Jest provides any way to mock `requestIdleCallback`

oliverlloyd:
> Would `deferUntilIdle` and `deferUntilVisible` send a stronger signal of relatedness?

I like having the word 'defer' in there. Maybe it helps to list all the potential deferations? The things I can imagine we might use to defer stuff are:

if idle
if visible
if media (some media query is true)
if mouse over (the mouseover event fires)
if mouse near (not sure how we achieve this but it's maybe nice to have for smaller components)

So, with your pattern, we could have

`deferUntilIdle`
`deferUntilVisible`
`deferUntilMedia`
`deferUntilMouseOver`
`deferUntilMouseNear`

with the 'when' construct it would be

`whenIdle`
`whenVisible`
`whenMedia`
`whenMouseOver`
`whenMouseNear`

I think both work. With defer it's a little more opinionated about the motivations for using the function but when is a bit more concise.

SiAdcock:
Thanks for laying all those examples out. I don't have strong opinions, but looking at the examples, `when` feels more like natural language, whereas `deferUntil` is a bit clunky (you need to think for a second) and noisy. Im happy to stick with `when`, and happy with this whole pattern 😊 Thanks @oliverlloyd!

SiAdcock (dismissed):

SiAdcock (commented):
Small thing: please could you update the [root README](https://github.com/guardian/libs/blob/main/README.md) to aid discoverability 🙏

src/whenIdle.test.ts
SiAdcock: Digging a bit more, [`@shopify/jest-dom-mocks`](https://github.com/Shopify/quilt/tree/main/packages/jest-dom-mocks) looks useful for features available in the browser. Currently, Coveralls is failing because [we're only testing the fallback case](https://coveralls.io/builds/44541082/source?filename=src%2FwhenIdle.ts#L17) when `requestIdleCallback` is unavailable.

oliverlloyd: 🤔 Given the only thing this abstraction really does is implement `requestIdleCallback`, do we risk a situation where we're purely testing the mock?

I guess there's an argument to say that our code is so simple anyway so the test will be the same and there's benefit in having them, if only for documentation?

SiAdcock: Yeah I think there's some benefit in knowing this function does what we expect, both with and without browser support.

oliverlloyd:
Do we want to resurrect this one or is everyone happy is I close this for now? Perhaps there's no use cases beyond DCR for this function, which is fine.

PR #297, opened 2021-11-23
feat: Added `whenVisible`

oliverlloyd:
## What?
`whenVisible` is a function that takes an element and a callback and calls the callback when the element is in the viewport

## Why?
This offers people a utility for deferring code execution.

This function is proposed in [this PR for React Islands](https://github.com/guardian/dotcom-rendering/pull/3629) but it is also something that could be useful in other areas

### DIFF
--- a/src/whenVisible.README.md
+++ b/src/whenVisible.README.md
@@ -14,4 +14,7 @@ const myElement = document.getElementById('my-element');
 whenVisible(myElement, function () {
     console.log('My element is now visible');
 });
+
+## Support for older browsers
+`IntersectionObserver` is (not supported)[https://caniuse.com/?search=IntersectionObserver] on IE or other older browsers. We check for support and fallback to returning early.
 ```
--- a/src/whenVisible.README.md
+++ b/src/whenVisible.README.md
@@ -14,7 +14,8 @@ const myElement = document.getElementById('my-element');
 whenVisible(myElement, function () {
     console.log('My element is now visible');
 });
+```
 
 ## Support for older browsers
+
 `IntersectionObserver` is (not supported)[https://caniuse.com/?search=IntersectionObserver] on IE or other older browsers. We check for support and fallback to returning early.
-```
--- a/src/whenVisible.README.md
+++ b/src/whenVisible.README.md
@@ -1,0 +1,15 @@
+# `whenIdle()`
+
+Returns: `void`
+
+Runs the given callback when it detects that the browser is idle or if 500ms has passed
+
+## Example
+
+```js
+import { whenIdle } from '@guardian/libs';
+
+whenIdle(function () {
+    console.log('The browser is idle');
+});
+```
--- a/src/whenVisible.README.md
+++ b/src/whenVisible.README.md
@@ -1,15 +1,17 @@
-# `whenIdle()`
+# `whenVisible()`
 
 Returns: `void`
 
-Runs the given callback when it detects that the browser is idle or if 500ms has passed
+Runs the given callback when it detects that the given element is visible
 
 ## Example
 
 ```js
-import { whenIdle } from '@guardian/libs';
+import { whenVisible } from '@guardian/libs';
 
-whenIdle(function () {
-    console.log('The browser is idle');
+const myElement = document.getElementById('my-element');
+
+whenVisible(myElement, function () {
+    console.log('My element is now visible');
 });
 ```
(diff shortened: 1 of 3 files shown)


src/whenVisible.README.md
```diff
@@ -0,0 +1,15 @@
+# `whenIdle()`
```
georgeblahblah, line 1: should this be for `whenVisible()` maybe?

oliverlloyd: Oops! (thanks 😄 )

src/whenVisible.ts
```diff
@@ -0,0 +1,20 @@
...
+): void => {
+	const io = new IntersectionObserver(([entry]) => {
+		if (!entry.isIntersecting) return;
+		// Disconnect this IntersectionObserver once seen
+		io.disconnect();
+		callback();
+	});
+
+	io.observe(marker);
+};
```
sndrs, line 20: could rename `marker` to `element`? it seems a more natural way of expressing it (e.g. it's used in the description) and it will pop up in intellisense etc

```suggestion
/**
 * Use this function to delay execution of something until an element is visible
 * in the viewport
 *
 * @param element : The html element that we want to observe;
 * @param callback : This is fired when the marker is visible in the viewport
 */
export const whenVisible = (
	element: HTMLElement,
	callback: () => void,
): void => {
	const io = new IntersectionObserver(([entry]) => {
		if (!entry.isIntersecting) return;
		// Disconnect this IntersectionObserver once seen
		io.disconnect();
		callback();
	});

	io.observe(element);
};
```

oliverlloyd: You're right, `element` is more semantic.

For some reason I never considered using suggested change over multiple lines... But this is perfect, thanks for this!

src/whenVisible.ts
```diff
@@ -0,0 +1,20 @@
...
+ * in the viewport
+ *
+ * @param element : The html element that we want to observe;
+ * @param callback : This is fired when the marker is visible in the viewport
+ */
+export const whenVisible = (
+	element: HTMLElement,
+	callback: () => void,
+): void => {
+	const io = new IntersectionObserver(([entry]) => {
```
SiAdcock, line 12: It's worth thinking about the experience in older browsers (IE11 😒), even if it's just to drop support for them entirely. The polyfill for IntersectionObserver is huge and certainly not worth bundling. It's likely applications that use polyfill.io will provide, or can be made to provide, IntersectiionObserver for older browsers. Should we add something to the README saying something along these lines?

src/whenVisible.ts
```diff
@@ -9,12 +9,17 @@ export const whenVisible = (
...
+			// Disconnect this IntersectionObserver once seen
+			io.disconnect();
+			callback();
+		});
 
-	io.observe(element);
+		io.observe(element);
+	} else {
+		// IntersectionObserver is not supported so failover to calling back immediately
+		callback();
```
mxdvl, line 23: Would you want to default to putting this at the end of the call stack?

```suggestion
		setTimeout(() => callback(), 0);
```

src/whenVisible.README.md
```diff
@@ -14,4 +14,7 @@ const myElement = document.getElementById('my-element');
 whenVisible(myElement, function () {
     console.log('My element is now visible');
 });
+
+## Support for older browsers
+`IntersectionObserver` is (not supported)[https://caniuse.com/?search=IntersectionObserver] on IE or other older browsers. We check for support and fallback to returning early.
 ```
```
mxdvl: Small formatting issue here, where you want the text outside of the three backticks.

src/whenVisible.ts
oliverlloyd: This is a great idea!

src/whenVisible.ts
```diff
@@ -0,0 +1,25 @@
...
+		const io = new IntersectionObserver(([entry]) => {
+			if (!entry.isIntersecting) return;
+			// Disconnect this IntersectionObserver once seen
+			io.disconnect();
+			callback();
+		});
+
+		io.observe(element);
+	} else {
+		// IntersectionObserver is not supported so failover to calling back immediately
```
mxdvl, line 22:
```suggestion
		// IntersectionObserver is not supported so failover to calling back at the end of the call stack
```

mxdvl (approved):
You’ve got my vote!

mxdvl (approved):

oliverlloyd:
Do we want to resurrect this one or is everyone happy is I close this for now? Perhaps there's no use cases beyond DCR for this function, which is fine.

PR #308, opened 2022-02-24
refactor: do some housekeeping

sndrs:
## What does this change?

- moves all the individual libs into their own directories to keep things a bit tidier
- uses our latest eslint configs
- updates our use of peerdeps to the lowest version in the peerdep's range (following https://github.com/guardian/source/pull/1215)
- bumps deps
- updates `logger#debug` to run whenever it's not on `https://www.theguardian.com`
  - addresses https://github.com/guardian/libs/security/code-scanning/2 

## Why?

it was getting a bit messy


sndrs:
@mxdvl is the change to logger ok with you?

src/logger/index.ts
```diff
@@ -30,11 +30,8 @@ const style = (team: Styles): string => {
  * Only logs in dev environments.
  */
 export const debug: LogCall = (team, ...args) => {
-	const isDevEnv =
-		window.location.host.includes('localhost') ||
-		window.location.host.endsWith('thegulocal.com') ||
-		window.location.host.endsWith('.dev-theguardian.com');
-	if (isDevEnv) log(team, ...args);
+	const isNotProd = window.location.origin !== 'https://www.theguardian.com';
```
sndrs, line 33: @mxdvl is this ok with you?

mxdvl (approved):
Fine by me.

Is there an assumption that this library runs only in a browser context? Probably for a separate PR.

package.json
```diff
@@ -23,44 +23,40 @@
...
+    "ts-node": "10.5.0",
     "tslib": "2.3.1",
-    "typescript": "4.3.5",
+    "typescript": "4.0.2",
     "wcag-contrast": "^3.0.0",
-    "web-vitals": "^2"
+    "web-vitals": "2.0.0"
+  },
+  "peerDependencies": {
+    "web-vitals": "2.*"
```
SiAdcock, line 59: As per [discussion here](https://github.com/guardian/source/pull/1215/files#r768142852) should we go with `^2.0.0` for consistency?

package.json
```diff
@@ -23,44 +23,40 @@
...
-    "rollup-plugin-terser": "7.0.2",
+    "prettier": "2.5.1",
     "semantic-release": "17.4.5",
-    "ts-jest": "27.0.5",
-    "ts-node": "10.2.1",
+    "ts-jest": "27.1.3",
+    "ts-node": "10.5.0",
     "tslib": "2.3.1",
-    "typescript": "4.3.5",
+    "typescript": "4.0.2",
```
SiAdcock, line 54: Is this an intentional downgrade of `typescript`?

package.json
```diff
@@ -23,44 +23,40 @@
...
     "wcag-contrast": "^3.0.0",
```
SiAdcock, line 55: Should we pin this version for consistency?

SiAdcock, line 55: Should we pin this version for consistency?

SiAdcock:
I agree with this refactor, it does indeed make things tidier 😄 

I wonder if, as we did with Source, we dispense with `index.ts` files under every feature directory and instead use filenames that reflect the name of the features (e.g. `src/cookies/cookies.ts`). 

We could also use the clever [`preserveModules` workaround](https://github.com/guardian/source/blob/0f64f51f9843107f1f539ddbe265e4f195cb107d/packages/%40guardian/source-react-components/rollup.config.js#L12-L25) to ensure unused features get tree-shaken.

Apologies if I've got the wrong of the stick on anything and I'm just adding noise to this PR 😅

package.json
sndrs: yeah same principle - make sure we're not using TS features that consumers may not have available. since it's also JS-compatible we can't define TS as a peerDep, but at least we can say it works with v4

sndrs:
> I wonder if, as we did with Source, we dispense with `index.ts` files under every feature directory and instead use filenames that reflect the name of the features (e.g. `src/cookies/cookies.ts`). 

👍 good idea. i might do in a separate PR though to keep things cleaner?

> We could also use the clever [`preserveModules` workaround](https://github.com/guardian/source/blob/0f64f51f9843107f1f539ddbe265e4f195cb107d/packages/%40guardian/source-react-components/rollup.config.js#L12-L25) to ensure unused features get tree-shaken.

this uses pkgu which still uses `tsc` to build, so no monolithic bundles here

SiAdcock (approved):
Lovely 😄

PR #314, opened 2022-03-02
feat: move all module exports to discrete files, to facilitate better tree-shaking

sndrs:
## What does this change?

breaks out module exports into the smallest files possible

## Why?

- this will help with tree-shaking in consumer builds
- core web vitals could do with better structure to enable this, so it’s largely left for now


mxdvl (approved):
This is brilliant!

Why is `yarn.lock` touched?

src/cookies/cookies.test.ts
```diff
@@ -1,5 +1,9 @@
 import MockDate from 'mockdate';
-import * as cookies from './index';
+import { getCookie } from './getCookie';
+import * as getCookieValues from './getCookieValues';
```
SiAdcock, line 3: Could we `import { getCookieValues }` here to make the import explicit? Or is this necessary to support jest's `spyOn` API?

SiAdcock (dismissed):
Awesomes! Let's do it! 👍

mxdvl (approved):

SiAdcock (approved):

PR #315, opened 2022-03-02
feat: add sticky_video to ophan component type

marjisound:
## What does this change?
This PR adds a new component STICKY_VIDEO and also adds two actions STICK & CLOSE to ophan types.
## Why?
In order to emit tracking event to ophan when a sticky video sticks to screen and when the sticky video is closed.

Another PR is created in ophan to add these to the thrift https://github.com/guardian/ophan/pull/4440

### DIFF
--- a/src/@types/ophan/index.ts
+++ b/src/@types/ophan/index.ts
@@ -38,7 +38,9 @@ type OphanAction =
 	| 'ACCEPT_DEFAULT_CONSENT'
 	| 'MANAGE_CONSENT'
 	| 'CONSENT_ACCEPT_ALL'
-	| 'CONSENT_REJECT_ALL';
+	| 'CONSENT_REJECT_ALL'
+	| 'STICK'
+	| 'CLOSE';
 
 type OphanComponentType =
 	| 'READERS_QUESTIONS_ATOM'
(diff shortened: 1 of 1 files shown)


sndrs (changes requested):
is this actually fixing something that was broken? it looks more like new functionality being added, which would be a new `feat` (if correct)

also, it's really helpful to explain in the PR title what's being affected. i assumed you were adding a sticky video component to libs, which def seemed the wrong place! but on closer inspection it's adding a new field that can be reported to ophan. the more information context a PR can provide the easier it is for reviewers to understand what the author is intending

sndrs:
not for this PR, but would it be easier to publish ophan types from ophan's codebase itself? the concomitant PR to ophan's thrift types will always need to be accompanied by one here i think? is it useful to separate them?

marjisound:
> not for this PR, but would it be easier to publish ophan types from ophan's codebase itself? the concomitant PR to ophan's thrift types will always need to be accompanied by one here i think? is it useful to separate them?

Thanks @sndrs for pointing it out. Do you think I should talk to Roberto about this?

marjisound:
> is this actually fixing something that was broken? it looks more like new functionality being added, which would be a new `feat` (if correct)
> 
> also, it's really helpful to explain in the PR title what's being affected. i assumed you were adding a sticky video component to libs, which def seemed the wrong place! but on closer inspection it's adding a new field that can be reported to ophan. the more information context a PR can provide the easier it is for reviewers to understand what the author is intending

Thanks @sndrs I changed the commit and PR message, do you think it's more descriptive of what it's doing now?

PR #320, opened 2022-03-09
docs: Add TS version to readme via an `update-readme` commit hook

sndrs:
## What does this change?

- adds an `update-readme` commit hook
- uses it to put the min TS version in the README	
- removes unpkg ref from docs (it's broken currently, and does anyone actually is it? doesn't sound like it...)

## Why?

information is power


SiAdcock (approved):
Swanky! 👢 📸 🤵

PR #323, opened 2022-03-25
feat: Record logs to data lake

mxdvl:
## What does this change?

Adds a method to record logs to our data lake. This offers a unified method for other teams to record data to the lake.

This uses Fastly’s logging tools and will surface data in BigQuery in near real-time.

## Why?

It already exists in at least three places:
- https://github.com/guardian/dotcom-rendering/blob/a0b9118abaa38d60a7004b804692fb11cb7da34f/dotcom-rendering/src/web/lib/logging.ts
- https://github.com/guardian/frontend/blob/80d199986debc65cf58320166575dc3138595a1e/static/src/javascripts/projects/commercial/sentinel.ts
- https://github.com/guardian/dotcom-rendering/blob/a0b9118abaa38d60a7004b804692fb11cb7da34f/dotcom-rendering/scripts/webpack/plugins/gu-stats-report-plugin.js


src/logger/record.ts
```diff
@@ -0,0 +1,112 @@
...
+type Data = {
+	label: string;
+	endpoint: string;
+	metrics?: Metrics;
+	properties?: Properties;
+};
+
+/**
+ * Record log to our Data Lake.
+ * ⚠️ Make sure the data collection is GDPR compliant.
```
mxdvl, line 71:
```suggestion
 * ⚠️ Make sure we have consent for collecting this data.
```

arelra (commented):
Great to see this in libs!

src/logger/record.ts
```diff
@@ -0,0 +1,117 @@
...
+ * - Numerical datum is of type `number`
+ *
+ * @param {Data} data The data to send
+ * @param {`${TeamName}.${string}`} data.label Used to identify the data (in BigQuery). Starts with a registered team name.
+ * @param {string} data.endpoint The endpoint to send the data to.
+ * @param {Properties} [data.properties] Nominal data. Defaults to an empty object.
+ * @param {Metrics} [data.metrics] Numerical data. Defaults to an empty object.
+ * @returns {boolean} Whether sending the data has been successfully queued.
+ */
+const recordLog = ({
```
kenoir, line 88: Can you weave a requirement for getting `ConsentState` for the CMP here, like the commercial-metrics lib does?

See: https://github.com/guardian/commercial-core/blob/main/src/send-commercial-metrics.ts#L157

That would help in making sure consent is respected.

kenoir: Worth noting that even if you respect consent choices here - because consent can be withdrawn retroactively (i.e. only the most recent consent choice matters), I think you'll need to have a filtered view on this table in the data-lake joining to a users current permissions if you want to properly respect consent.

mchv:
We should not record undocumented and untyped data to our data warehouse.
We have moved from a data lake to a data warehouse where we expect:
- [Data minimisation ](https://edps.europa.eu/data-protection/data-protection/glossary/d_en)and privacy by design principle
- Collect only relevant data after we have agreed with the business data owners and lifecycle
- Document every data collected at source so it can be correctly transformed and aggregated in our data platform.

mxdvl:
I have taken another crack at this to make it as explicit as possible that this should only be used to capture anonymous data. I hope this addresses some of the concerns of @kenoir & @mchv.

I believe there is a real need for this kind of logs, when we do not need to join them to page views.

kenoir:
Sorry for the late reply, @dominickendrick is now the EM for Transparency & Consent.

I think this is a really hard question as even if we add these checks for `pageViewId` & `browserId` it's still possible that free text logs contain identifying information, however that is also true for existing logging solutions & looking at the examples of where we already do this in DCR & frontend _already in place_ without even these basic checks.

This PR in conjunction with updates that move the existing examples to use this lib would seem to materially improve the situation as it exists now.

@mchv I think we should either move towards consolidating the multiple existing places where we are pushing logs to the data lake with clearly defined responsibility for the consolidated solution (with this change as a step in that direction) or take a decision to deprecate and remove the existing examples of logging to BigQuery from the codebase.

sndrs:
@mxdvl my feeling is we should close this for now?

@kenoir's final comment makes a lot of sense, and libs would be a good place for it, but it's not clear the business is ok with it. i think formalising something in this package implies a greater degree org-wide agreement than we actually have. 

it's shame because this would be really useful...

PR #327, opened 2022-04-05
fix: time separator is `.`

mxdvl:
## What does this change?

Fixes the time separator for the `withTime` method.

## Why?

From [the Guardian and Observer style guide: T][T]

> **times**
> 1am, 6.30pm, etc; 10 o’clock last night but 10pm yesterday; half past two, a quarter to three, 10 to 11, etc; 2hr 5min 6sec, etc; for 24-hour clock, 00.47, 23.59; noon, midnight (not 12 noon, 12 midnight or 12am, 12pm).

[T]: https://www.theguardian.com/guardian-observer-style-guide-t

### DIFF
--- a/src/datetime/timeAgo.test.ts
+++ b/src/datetime/timeAgo.test.ts
@@ -72,7 +72,7 @@ describe('timeAgo', () => {
 		const yesterday = new Date(Date.UTC(2019, 10, 16, 3, 0, 0)).getTime();
 
 		expect(timeAgo(yesterday)).toBe('1d ago');
-		expect(timeAgo(yesterday, { verbose: true })).toBe('Yesterday 3:00');
+		expect(timeAgo(yesterday, { verbose: true })).toBe('Yesterday 3.00');
 	});
 
 	it('does not pluralise the unit when the delta is one', () => {
@@ -101,7 +101,7 @@ describe('timeAgo', () => {
 			timeAgo(oneDayAgo, {
 				verbose: true,
 			}),
-		).toBe('Yesterday 12:00');
+		).toBe('Yesterday 12.00');
 	});
 
 	it('returns verbose format for seconds when this option is given', () => {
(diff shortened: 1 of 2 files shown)


src/datetime/timeAgo.test.ts
```diff
@@ -72,7 +72,7 @@ describe('timeAgo', () => {
 		const yesterday = new Date(Date.UTC(2019, 10, 16, 3, 0, 0)).getTime();
 
 		expect(timeAgo(yesterday)).toBe('1d ago');
-		expect(timeAgo(yesterday, { verbose: true })).toBe('Yesterday 3:00');
+		expect(timeAgo(yesterday, { verbose: true })).toBe('Yesterday 3.00');
```
mxdvl, line 75: The style guide seems to suggest this should omit the minutes and opt for `Yesterday 3`

src/datetime/timeAgo.ts
```diff
@@ -19,7 +19,7 @@ const shortMonth = (month: number): string =>
 const longMonth = (month: number): string =>
 	[
 		'January',
-		'Febuary',
+		'February',
```
sndrs, line 22: 😱

src/datetime/timeAgo.test.ts
sndrs: i'd be interested to know what UX think. intuitively it feels like we have multiple use cases here. a journalist writing "yesterday at 1am" seems very different to a status of "15.00". the `.` though def seems to fit the style guide.

sndrs (commented):
do you want to add reference and link to the style guide at the top of the file too?

src/datetime/timeAgo.ts
mxdvl: It’s how people say it, innit, though?

sndrs (approved):

PR #328, opened 2022-04-05
refactor(getCookie): get the value directly

mxdvl:
## What does this change?

Refactor the map check in getCookie to always `get`, instead of checking `has` before.

## Why?

Currently, we don’t have 100% coverage because of [a TS limitation](https://github.com/microsoft/TypeScript/issues/9619#issuecomment-473669366).
There’s safety built-in the get method, so let’s use it directly.

We could alternatively use a `//@ts-expect-error` or an `as string` type assertion.


mxdvl:
Duplicate of #335

PR #334, opened 2022-04-19
refactor(Format): use objects with as const assertions

mxdvl:
## What does this change?

Turn the `Format` enums to simple objects. Using the `as const` assertion gives the same level of type-safety as enums.

## Why?

There are strange strange quirks with typescript enums, which yield an object with both keys as keys and keys as values.

For example, `enum { A, B }` yields:

```
{
 A: 0,
 B: 1,
 "0": "A",
 "1": "B",
}
```


src/cookies/getCookie.ts
```diff
@@ -15,9 +15,8 @@ export const getCookie = ({
 	name: string;
 	shouldMemoize?: boolean;
 }): string | null => {
-	if (memoizedCookies.has(name)) {
-		return memoizedCookies.get(name) ?? null;
-	}
+	const memoizedCookie = memoizedCookies.get(name);
+	if (memoizedCookie) return memoizedCookie;
```
sndrs, line 19: is this from an older PR? maybe #335

mxdvl: Ah yes—an issue with squash and merge!

mxdvl:
Not sure if this is needed yet… might revisit in the future.

PR #335, opened 2022-04-19
refactor(getCookie): retrieve memoized directly

mxdvl:
## What does this change?

There’s no need to check if the value exists before retrieving it, as the default behaviour of `get` on a Map is to return undefined if the key is missing. If the key exists, return it directly.

## Why?

This allows us to reach true 100% coverage, as the missing code path was unreachable.


src/cookies/getCookie.ts
```diff
@@ -15,9 +15,8 @@ export const getCookie = ({
 	name: string;
 	shouldMemoize?: boolean;
 }): string | null => {
-	if (memoizedCookies.has(name)) {
-		return memoizedCookies.get(name) ?? null;
```
sndrs, line 19: this would be a breaking change in theory (returning `null` instead of `undefined`). is anything already using this explicitly check for `null`? if not we can probably get away with it, but i think the type signature should also be updated?

sndrs (changes requested):

src/cookies/getCookie.ts
mxdvl: If you’re in this code block, the value is not `undefined`, because we checked that is exists with `.has(name)`. By switching to using `.get(name)` directly, we only return the value if it’s not undefined.

The reason we had the nullish coalescing was present is only because TypeScript was unable to discriminate a map that has a value.

sndrs (approved):

PR #336, opened 2022-04-19
feat: as isEnum method

mxdvl:
## What does this change?

Add a new `isEnum` method to check is an object has a unique set of keys and values.

## Why?

We’re proposing objects with `as const` assertions instead of enums in #334


mxdvl:
Might revisit in the future…

PR #349, opened 2022-05-10
revert(ophan)!: Revert add `key_event_card` ophan component type

joecowton1:
Reverts guardian/libs#348

We no longer need this change, see this comment: https://github.com/guardian/dotcom-rendering/pull/4856#pullrequestreview-967516855


sndrs:
this would be a breaking change i think? you can do `fix!` to trigger that I think

joecowton1:
> this would be a breaking change i think? you can do `fix!` to trigger that I think

Ah yes you're right, should be fixed now.

mxdvl:
Alternatively, mark it as deprecated, and remove it in another version, if you don't need this to trigger a major release.

joecowton1:
> Alternatively, mark it as deprecated, and remove it in another version, if you don't need this to trigger a major release.

Happy to do either, though I imagine it may get forgotten if we don't remove it now. What is your preference @sndrs ?

jamesgorrie (approved):

jamesgorrie:
I'd go with triggering a major release - any changes in the client will be simple enough and avoids creating tech debt.

sndrs:
agree with @jamesgorrie – let's own this change :)

sndrs (approved):

PR #376, opened 2022-07-29
refactor!: Change `format` from enum to string type

DanielCliftonGuardian:
## What does this change?

This changes the format types from enums to string.

## Why?

This helps bring DCR in line with the Frontend string properties reducing the number of transformations.

Secondly this will overcome difficulties of working with enums such as here:

https://github.com/guardian/dotcom-rendering/blob/main/dotcom-rendering/scripts/gen-stories/get-stories.js#L119


src/format/ArticleDesign.ts
```diff
@@ -1,25 +1,24 @@
...
-	PhotoEssay,
-	PrintShop,
-	Obituary,
-	Correction,
-	FullPageInteractive,
-	NewsletterSignup,
-}
+type ArticleDesign =
+	| 'DesignStandard'
+	| 'DesignMedia'
```
JamieB-gu, line 3: I think this would be "Gallery", "Audio" and "Video" rather than "Media"?

JamieB-gu (commented):
Interesting! Would you mind talking a little bit more about why this change is being made? It will mean some fairly major refactoring downstream in DCR, AR, atoms-rendering (I think those are all the places we use `Format`?), so would be good to understand the motivations. I'm not suggesting it's wrong btw, would just like to understand a bit more 🙂.

Also, could I suggest dropping the `Design`, `Display`, `Pillar` and `Special` prefixes on each of the values? So we have something like:

```ts
type ArticleDesign =
  | 'Standard'
  | 'Gallery'
  | 'Review'
  ...
```

and so on. I think we already know this information from the name of the type itself, and it would probably make these easier to read and reduce string size in the bundle?

src/format/ArticleSpecial.ts
```diff
@@ -1,4 +1,3 @@
-export enum ArticleSpecial {
-	SpecialReport = 5,
-	Labs = 6,
-}
+type ArticleSpecial = 'SpecialReport' | 'SpecialLabs';
```
oliverlloyd, line 1: I'm not 100% sure how up to date [these Frontend types](https://github.com/guardian/frontend/blob/oliver/press-aus-weekend/common/app/model/meta.scala#L288) are but do we want to follow the pattern there and have `SpecialReportTheme` and then maybe `Labs` (or perhaps `LabsTheme`?).

oliverlloyd (commented):
This was a piece of work that came out of the Dotcom dev time

I paired on this so I'll leave it to cs infra to officially approve but to my eyes this looks good.

Do we need to update the README a bit further so the example is always importing a `type`?

jamesgorrie (commented):
This does feel like it would be much clearer and easier to debug. Nice!

Given part of what this is helping solve (from the example) is making it easier to iterate on these, will we have to create them as a value as well as a type?

The suggestion @mxdvl and I worked on that allowed this was something like:
```TS
const ArticleDesigns = ['Standard', 'Gallery', 'Review'] as const;
type ArticleDesign = typeof ArticleDesigns[number];

for (const design in ArticleDesigns) {
    console.info(design)
}
```

src/format/ArticleSpecial.ts
bryophyta: @OllysCoding, when this came up when we were working on formats-explorer, I think you had some thoughts relating to these categories?

DanielCliftonGuardian:
> This does feel like it would be much clearer and easier to debug. Nice!
> 
> Given part of what this is helping solve (from the example) is making it easier to iterate on these, will we have to create them as a value as well as a type?
> 
> The suggestion @mxdvl and I worked on that allowed this was something like:
> 
> ```ts
> const ArticleDesigns = ['Standard', 'Gallery', 'Review'] as const;
> type ArticleDesign = typeof ArticleDesigns[number];
> 
> for (const design in ArticleDesigns) {
>     console.info(design)
> }
> ```

I have implemented this suggestion. Thanks

joecowton1:
> I paired on this so I'll leave it to cs infra to officially approve but to my eyes this looks good.

This seems like an interesting change, we discussed in CSTI planning and have no obvious objections but we're cautious these types are used are in a number of places - commercial-core / source / atoms-rendering / apps-rendering / image-rendering / discussion-rendering (and maybe more) so would require a decent amount of dev time to bring everything up to date. 

Could you provide a little more context as to why we should do it and what the initial motivation was for looking into it please @DanielCliftonGuardian ? Thanks!

DanielCliftonGuardian:
> > I paired on this so I'll leave it to cs infra to officially approve but to my eyes this looks good.
> 
> This seems like an interesting change, we discussed in CSTI planning and have no obvious objections but we're cautious these types are used are in a number of places - commercial-core / source / atoms-rendering / apps-rendering / image-rendering / discussion-rendering (and maybe more) so would require a decent amount of dev time to bring everything up to date.
> 
> Could you provide a little more context as to why we should do it and what the initial motivation was for looking into it please @DanielCliftonGuardian ? Thanks!
> 
> @sndrs @tjsilver @ashishpuliyel



> > I paired on this so I'll leave it to cs infra to officially approve but to my eyes this looks good.
> 
> This seems like an interesting change, we discussed in CSTI planning and have no obvious objections but we're cautious these types are used are in a number of places - commercial-core / source / atoms-rendering / apps-rendering / image-rendering / discussion-rendering (and maybe more) so would require a decent amount of dev time to bring everything up to date.
> 
> Could you provide a little more context as to why we should do it and what the initial motivation was for looking into it please @DanielCliftonGuardian ? Thanks!
> 
> @sndrs @tjsilver @ashishpuliyel

Hi @joecowton1, It might be better to refer to @oliverlloyd whom i paired with on this for a deeper insight into the benefits of this change as I am very new to the team and don't have much more to add.

sndrs:
hi @DanielCliftonGuardian – is this still needed? i don't think i can i really offer a review of whether we should do it, but the changes look like they should work!

is everyone good for this to be merged?

mxdvl:
@DanielCliftonGuardian is on holiday for a while – I think one way forward with this would be to bring it up at the Client-Side infra meeting where most consumers are likely to get visibility. In the meantime it might be good to publish this as a pre-release so we can test how big a change it requires on consumers?

JamieB-gu:
> is everyone good for this to be merged?

I think the points I raised in my comments above probably still need looking into?

oliverlloyd:
+1 to @JamieB-gu 's [points above](https://github.com/guardian/libs/pull/376#pullrequestreview-1055713089). It'd be great to drop the prefixes and have something a little nicer to read and we need to have support for Audio, Video, etc.,.

I also think @mxdvl 's idea of raising this in the client fireside chat is a good idea.

sndrs:
i'm hoping to move libs to https://github.com/guardian/csnx in the next few days, hence asking about this PR. 

I'll leave it open for now, but the repo will be archived soon. 

hopefully leaving this PR open will capture the thoughts already in here in a useful way for posterity, and the changes themselves are easy enough to recreate.

does that sound ok with everyone?

oliverlloyd:
> the changes themselves are easy enough to recreate

I think this is true 👍

Issue #350, opened 2022-05-11
should we move Core Web Vitals to their own package?


sndrs:
main reason being it turns the google package into a peerDep of libs regardless of what you're using. 

i'm getting warning about it being missing when using libs in another node module. 

i don't want to add it to keep that warning quiet.

maybe anything that has a peerDep should not be included in libs?

any thoughts? 

cc @mxdvl

mxdvl:
I think that’s a valid point, and it would be better to keep `@guardian/libs` dependency-free. Any idea where it might live instead?

I’ve been thinking recently about creating a better abstraction around this anyways, and we could possibly move the core-vitals dependencies back to the consumers. Something like below, where the `init` function returns a method you can call to register sending the metrics at page unload:

```ts
/** List of collecting functions to run at page unload */
const collectors = new Map<string, () => void>();

/** Initialiser logger */
const init = ({ label, collector }: {
  label: string;
  collector: () => void;
}) => ({
  /** Call this function to schedule collection at page unload */
  schedule: (consent: boolean) => {
    if (!consent) {
      console.warn(`No consent for “${label}” logging`);
      return;
    }

    if (collectors.has(label)) {
      console.warn(`Overriding “${label}” logging`);
    }

    collectors.set(label, collector);
  },
  /** Call this function to stop collection at page unload */
  remove: () => {
    console.warn(`Removing “${label}” logging`);
    collectors.delete(label);
  },
});
```

PR #377, opened 2022-07-29
chore(codeowners): remove old, add new CODEOWNERS

JamieB-gu:
## What does this change?

The CODEOWNERS file was showing errors, fixed some of them in the following ways.

- `gtrufitt` no longer has access, removed from the file.
- `guardian-frontend-team` doesn't seem to exist any more, removed from the file. Added `dotcom-platform` instead.

Added `apps-rendering` and `dotcom-platform` as owners of the `format` module.

## Why?

GitHub was showing three errors in the CODEOWNERS file. I've fixed two of them as described above. There's still a problem with `commercial-dev`, and I can see a similar error has appeared for the new teams I've added. Perhaps we need to update the repo permissions for this to work?

I've made the `format` ownership change because I think the Apps-Rendering and Dotcom teams probably own this module and any downstream repos that will be affected by changes to it (DCR, AR, atoms-rendering)? Please let me know if this is an incorrect assumption!

**Note:** I've assumed `dotcom-platform` is a valid replacement for the `guardian-frontend-team` as an owner of the core web vitals module. If this is wrong please let me know!

### DIFF
--- a/.github/CODEOWNERS
+++ b/.github/CODEOWNERS
@@ -1,6 +1,8 @@
 # if you would like to help maintain this repo,
 # please add yourself in a PR
 
-* @guardian/client-side-infra @gtrufitt
+* @guardian/client-side-infra
 
-src/coreWebVitals*  @guardian/guardian-frontend-team @guardian/commercial-dev
+src/coreWebVitals* @guardian/dotcom-platform @guardian/commercial-dev
+
+/src/format/ @guardian/apps-rendering @guardian/dotcom-platform


jamesgorrie:
>  I can see a similar error has appeared for the new teams I've added

jamesgorrie (commented):
>  I can see a similar error has appeared for the new teams I've added

> The people you choose as code owners must have write permissions for the repository. When the code owner is a team, that team must be visible and it must have write permissions, even if all the individual members of the team already have write permissions directly, through organization membership, or through another team membership.

- https://docs.github.com/en/repositories/managing-your-repositorys-settings-and-features/customizing-your-repository/about-code-owners

If we think this is the right thing, then we should add them.

A question I have, is why are we the codeowners or this specific piece?

sndrs:
Apologies for missing this...

> A question I have, is why are we the codeowners or this specific piece?

yeah that's a good question. another way of looking at it: if it's only dotcom/ar who need `format`, should it really live in here?

sndrs:
re `src/coreWebVitals`, i think i'd rather resolve https://github.com/guardian/libs/issues/350 (and probably _not_ keep `coreWebVitals` in here)

oliverlloyd:
> if it's only dotcom/ar who need `format`,

That's true at the moment but this could change?

If we did remove it from /libs then are we saying we'd inline it back into AR and DCR?

sndrs:
i've added the teams tot he repo – shall we merge this as is and revisit the questions in different hcanges?

oliverlloyd:
It would be good to merge this so that we can unblock https://github.com/guardian/libs/pull/386 - which is on the critical path for some changes which have a lot of focus at the moment

sndrs (approved):

JamieB-gu:
> another way of looking at it: if it's only dotcom/ar who need `format`, should it really live in here?

Perhaps AR/Dotcom could be considered maintainers, but there's nothing stopping other teams from contributing to and using it? In the same way that CS Infra maintains Source but other teams use and contribute to it? And AR and Dotcom maintain AR and DCR respectively, but other teams use and contribute to those platforms?

In all those cases, you'd probably want the maintainers to be flagged for automatic review, regardless of who opens a PR to make changes. For example, #386 as @oliverlloyd mentions, where Journalism are making a change to `Format` - we'd have AR and Dotcom review that change.

PR #383, opened 2022-08-04
fix: Remove `any` types

mxdvl:
## What does this change?

Remove `any` types from the `storage` API.

## Why?

This caused an issue in Production recently, where the type that became returned as a `number` instead of a string.


ob6160 (approved):
great idea, makes sense to me 👍

src/storage/storage.ts
```diff
@@ -1,3 +1,6 @@
+import { isObject } from '../isObject/isObject';
+import { isString } from '../isString/isString';
```
mxdvl, line 2: Is is okay to import helpers from other parts of libs?

ob6160: the same pattern seems to be used in other places: https://github.com/guardian/libs/blob/1912df53e54ee231539353746d554287a7e59013/src/locale/getLocale.ts#L4

sndrs (approved):

PR #385, opened 2022-08-09
feat(logger): add `identity` team to logger

coldlink:
## What does this change?

Add the Identity team to the logger

## Why?

So the Identity team can use these logs to identify their own logs in other platforms, e.g. frontend/dcr, as well as our own.

### DIFF
--- a/src/logger/teamStyles.ts
+++ b/src/logger/teamStyles.ts
@@ -33,4 +33,8 @@ export const teamStyles = {
 		background: '#0F70B7',
 		font: '#ffffff',
 	},
+	identity: {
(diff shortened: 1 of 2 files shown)


sndrs:
i think this is a `feat` rather than a `fix`? all good apart from that

sndrs (commented):
i think [this is a `feat` rather than a `fix`](https://semver.org/#summary)? all good apart from that

coldlink:
Thanks! That makes sense, updated commit/pr to `feat`

sndrs (approved):

PR #390, opened 2022-08-23
fix: Use `@guardian/tsconfig`

sndrs:
## What does this change?

Adds `@guardian/tsconfig` to the typscript config

## Why?

It's stricter (in a good way) and preps the code for import to https://github.com/guardian/csnx

### DIFF
--- a/src/cookies/getCookie.ts
+++ b/src/cookies/getCookie.ts
@@ -18,7 +18,7 @@ export const getCookie = ({
 	const memoizedCookie = memoizedCookies.get(name);
 	if (memoizedCookie) return memoizedCookie;
 
-	const value = getCookieValues(name)[0];
+	const [value] = getCookieValues(name);
 
 	if (value) {
 		if (shouldMemoize) {
--- a/src/cookies/getCookie.ts
+++ b/src/cookies/getCookie.ts
@@ -20,7 +20,7 @@ export const getCookie = ({
 
 	const cookieVal = getCookieValues(name);
 
-	if (cookieVal.length > 0) {
+	if (cookieVal[0]) {
 		if (shouldMemoize) {
 			memoizedCookies.set(name, cookieVal[0]);
 		}
--- a/src/cookies/getCookie.ts
+++ b/src/cookies/getCookie.ts
@@ -20,10 +20,8 @@ export const getCookie = ({
 
 	const cookieVal = getCookieValues(name);
 
-	if (cookieVal[0]) {
-		if (shouldMemoize) {
-			memoizedCookies.set(name, cookieVal[0]);
-		}
+	if (shouldMemoize && cookieVal[0]) {
+		memoizedCookies.set(name, cookieVal[0]);
 		return cookieVal[0];
 	}
 	return null;
--- a/src/cookies/getCookie.ts
+++ b/src/cookies/getCookie.ts
@@ -18,11 +18,13 @@ export const getCookie = ({
 	const memoizedCookie = memoizedCookies.get(name);
 	if (memoizedCookie) return memoizedCookie;
 
-	const cookieVal = getCookieValues(name);
+	const value = getCookieValues(name)[0];
 
-	if (shouldMemoize && cookieVal[0]) {
-		memoizedCookies.set(name, cookieVal[0]);
-		return cookieVal[0];
+	if (value) {
+		if (shouldMemoize) {
+			memoizedCookies.set(name, value);
+		}
+		return value;
 	}
 	return null;
 };
--- a/src/cookies/setCookie.ts
+++ b/src/cookies/setCookie.ts
@@ -45,6 +45,9 @@ export const setCookie = ({
 
 	// If the cookie is already memoized we want to replace its value
 	if (memoizedCookies.has(name)) {
-		memoizedCookies.set(name, getCookieValues(name)[0]);
+		const value = getCookieValues(name)[0];
+		if (value) {
+			memoizedCookies.set(name, value);
+		}
 	}
 };
--- a/src/cookies/setCookie.ts
+++ b/src/cookies/setCookie.ts
@@ -45,7 +45,7 @@ export const setCookie = ({
 
 	// If the cookie is already memoized we want to replace its value
 	if (memoizedCookies.has(name)) {
-		const value = getCookieValues(name)[0];
+		const [value] = getCookieValues(name);
 		if (value) {
 			memoizedCookies.set(name, value);
 		}
--- a/src/cookies/setSessionCookie.ts
+++ b/src/cookies/setSessionCookie.ts
@@ -26,7 +26,7 @@ export const setSessionCookie = ({
 
 	// If the cookie is already memoized we want to replace its value
 	if (memoizedCookies.has(name)) {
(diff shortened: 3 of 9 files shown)


ob6160 (dismissed):
nice work 👍

src/cookies/getCookie.ts
```diff
@@ -20,7 +20,7 @@ export const getCookie = ({
 
 	const cookieVal = getCookieValues(name);
 
-	if (cookieVal.length > 0) {
+	if (cookieVal[0]) {
```
ob6160: i wonder if we could shift the array access up to `const cookieVal = getCookieValues(name)[0];` for consistency with https://github.com/guardian/libs/pull/390/files#diff-fed7a7bee596fcd23f90f5770a22a3d2828ce9d22c8a0f46518c6c6715798498R48 for example?

src/datetime/timeAgo.ts
```diff
@@ -137,7 +105,9 @@ export const timeAgo = (
 		// Simple date - "9 Nov 2019"
 		return [
 			then.getDate(),
-			verbose ? longMonth(then.getMonth()) : shortMonth(then.getMonth()),
+			verbose
```
ob6160, line 108: ah! a lot nicer :)

src/datetime/timeAgo.ts
```diff
@@ -137,7 +105,9 @@ export const timeAgo = (
...
+				? then.toLocaleString('default', { month: 'long' })
+				: then.toLocaleString('default', { month: 'short' }),
```
sndrs, line 110: @oliverlloyd how do you feel about these? guardian is at 98.99% support (greater than es module support). are they going to have wired side effects?

JamieB-gu: Given that the other functions in this file return English strings, and the style guide is British English, is it possible we want to use `'en-GB'` here?

```suggestion
				? then.toLocaleString('en-GB', { month: 'long' })
				: then.toLocaleString('en-GB', { month: 'short' }),
```

I think using `'default'` will use the browser/system's default locale, and will therefore give the month in the language for that locale? For example, in French if the system is set to French. You can test this with the following snippet of code:

```js
const d = new Date();

d.toLocaleString('default', { month: 'long' });
```

in Node by setting the `LANG` environment variable:

```sh
LANG=fr_FR node
```

or in Safari by changing the macOS system language to e.g. French, restarting your browser, and then pulling up a console.

src/cookies/getCookie.ts
```diff
@@ -18,13 +18,13 @@ export const getCookie = ({
 	const memoizedCookie = memoizedCookies.get(name);
 	if (memoizedCookie) return memoizedCookie;
 
-	const cookieVal = getCookieValues(name);
+	const value = getCookieValues(name)[0];
```
mxdvl, line 21:
```suggestion
	const [value] = getCookieValues(name);
```

src/cookies/setCookie.ts
```diff
@@ -45,6 +45,9 @@ export const setCookie = ({
 
 	// If the cookie is already memoized we want to replace its value
 	if (memoizedCookies.has(name)) {
-		memoizedCookies.set(name, getCookieValues(name)[0]);
+		const value = getCookieValues(name)[0];
```
mxdvl, line 48:
```suggestion
		const [value] = getCookieValues(name);
```

src/cookies/setSessionCookie.ts
```diff
@@ -26,6 +26,9 @@ export const setSessionCookie = ({
 
 	// If the cookie is already memoized we want to replace its value
 	if (memoizedCookies.has(name)) {
-		memoizedCookies.set(name, getCookieValues(name)[0]);
+		const value = getCookieValues(name)[0];
```
mxdvl, line 29:
```suggestion
		const [value] = getCookieValues(name);
```

src/datetime/timeAgo.ts
mxdvl: I would be tempted to move this to a separate PR, as it’s possibly a breaking change, or at least most definitly a feature.

My suggestion for the existing code would be:

```ts
const longMonth = (month: number): string =>
	([
		'January',
		'February',
		'March',
		'April',
		'May',
		'June',
		'July',
		'August',
		'September',
		'October',
		'November',
		'December',
	] as const)[month] ?? "Unknown";
	
const shortMonth = (month: number) => longMonth(month).slice(0,3);
```

tsconfig.json
```diff
@@ -1,18 +1,19 @@
...
-  "exclude": ["node_modules", "dist", "coverage"],
-  "ts-node": { "compilerOptions": { "module": "commonjs" } }
+	"extends": "@guardian/tsconfig",
+	"compilerOptions": {
+		"allowJs": true,
+		"allowSyntheticDefaultImports": true,
+		"esModuleInterop": true,
+		"forceConsistentCasingInFileNames": true,
+		"isolatedModules": true,
+		"module": "ESNext",
```
mxdvl, line 9: I think you can remove `allowSyntheticDefaultImports`, `esModuleInterop` and `module` and `moduleResolution`.

mxdvl (approved):
This is great, but I would not change the behaviour of `timeAgo` as part of this change.

src/datetime/timeAgo.ts
sndrs: this is a great point – `'en-GB'` should resolve that though?

sndrs:
> I don’t think this should go in as-is, because of the timeAgo changes.

what do you think needs to change to make them ok? the `?? "Unknown"` is only there because as far as i know TS doesn't have a type for `0-11`. that makes it a branch of code we can't test, and given the month index comes from `getMonth`, a branch we'll never use. it's only there because `getMonth` returns a `number` in ts.

if the `en-GB` @JamieB-gu mentioned returns the same strings, shouldn't we be ok? what would you think we'd need to address in a subsequent PR?

src/datetime/timeAgo.ts
oliverlloyd: I think `then.toLocaleString('en-GB', { month: 'short' })` best respects the design guide

mxdvl: I thought that [`toLocaleString` was not well supported in browsers](https://caniuse.com/mdn-javascript_builtins_date_tolocalestring), but it seems like that’s not the case, there’s only missing data for about 3% of browsers.

mxdvl (approved):
Indeed, with the great support of `toLocaleString`, I think this is good to go!

src/datetime/timeAgo.ts
sndrs: Support is available for everything that supports es modules afaict?

JamieB-gu:
@sndrs did we want to change the `toLocaleString` call to specify `en-GB`? I think the code now on `main` will show datetimes in different languages if any of the consumers update to it? I may have missed something though 🤔

sndrs:
Doh! Yes I completely forgot to implement it!

sndrs:
@JamieB-gu correct locale went out in https://github.com/guardian/libs/releases/tag/v7.1.3

PR #391, opened 2022-08-24
perf!: Import helpers from `tslib`

sndrs:
## What does this change?

stops bundling helpers from `tslib` in the published code

## Why?

consumers can publish smaller bundles

## Before

```js
var __classPrivateFieldSet = (this && this.__classPrivateFieldSet) || function (receiver, privateMap, value) {
    if (!privateMap.has(receiver)) {
        throw new TypeError("attempted to set private field on non-instance");
    }
    privateMap.set(receiver, value);
    return value;
};
var __classPrivateFieldGet = (this && this.__classPrivateFieldGet) || function (receiver, privateMap) {
    if (!privateMap.has(receiver)) {
        throw new TypeError("attempted to get private field on non-instance");
    }
    return privateMap.get(receiver);
};
```

## After

```js
import { __classPrivateFieldGet, __classPrivateFieldSet } from "tslib";
```


ob6160 (dismissed):

mxdvl:
What happened with the whitespace changes?

sndrs:
> What happened with the whitespace changes?

wow i really don't know! sorry about that, should be fixed now

ob6160 (approved):

